PR #11, opened 2017-04-18
handling for gradle buildfiles in different dirs

joschli:


mubench.pipeline/tests/tasks/implementations/test_compile.py
```diff
@@ -294,7 +294,41 @@ def test_compile_with_gradle(self, shell_mock, copy_mock):
...
+
+    @patch("tasks.implementations.compile.shutil.copy")
+    @patch("tasks.implementations.compile.Shell.exec")
+    def test_compile_with_gradle_project_dir_parameter(self, shell_mock, copy_mock):
+        self.version._YAML["build"]["commands"] = ["gradle build -p /testdir/"]
+        shell_mock.return_value = ":printClasspath\n/path/dependency1.jar\n/path/dependency2.jar\n\nBUILD SUCCESSFUL"
+
+        self.uut.process_project_version(self.project, self.version)
+        assert_equals(shell_mock.mock_calls[1][1], ("gradle :printClasspath -b '/testdir/classpath.gradle'",))
+        assert_in(call("/path/dependency1.jar", self.dep_path), copy_mock.mock_calls)
```
salsolatragus: test_compile_with_gradle already checks that parsing the classpath from the output works. No need to repeat this in every subsequent test. Please remove the respective asserts (and the mock setup, if possible).

mubench.pipeline/tests/tasks/implementations/test_compile.py
```diff
@@ -294,7 +294,41 @@ def test_compile_with_gradle(self, shell_mock, copy_mock):
...
+        shell_mock.return_value = ":printClasspath\n/path/dependency1.jar\n/path/dependency2.jar\n\nBUILD SUCCESSFUL"
+
+        self.uut.process_project_version(self.project, self.version)
+        assert_equals(shell_mock.mock_calls[1][1], ("gradle :printClasspath -b '/test dir/classpath.gradle'",))
+        assert_in(call("/path/dependency1.jar", self.dep_path), copy_mock.mock_calls)
+        assert_in(call("/path/dependency2.jar", self.dep_path), copy_mock.mock_calls)
+
+    @patch("tasks.implementations.compile.shutil.copy")
+    @patch("tasks.implementations.compile.Shell.exec")
+    def test_compile_with_gradle_project_dir_parameter_spaces2(self, shell_mock, copy_mock):
```
salsolatragus: The difference between the spaces/spaces2 tests is single vs. double quotes, right? Please change the test names to reflect that, e.g., test_compile_with_gradle_project_dir_parameter_space_single_quotes.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+
+            # add rest of dir name if not even amount of " or ' in dir name (dirs with spaces are wrapped)
+            while buildfile_dir.count('"') % 2 is not 0 or buildfile_dir.count('\'') % 2 is not 0:
+                last_index += 1
+                buildfile_dir += " " + args[last_index]
+
+        # remove ' or " and add / if missing
+        buildfile_dir = buildfile_dir.replace("'", "")
+        buildfile_dir = buildfile_dir.replace("\"", "")
+        if buildfile_dir is not "" and not buildfile_dir.endswith("/"):
```
salsolatragus: use os.path.join(buildfile_dir, "classpath.gradle") to obtain the path. Makes the check unnecessary.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+            buildfile_dir += "/"
+
+        cmd = "gradle :printClasspath -b '" + buildfile_dir + "classpath.gradle'"
```
salsolatragus: Use "gradle :printClasspath -b '{}'".format(buildfile_path).
Is there a difference to "gradle ... -p {} -b classpath.gradle".format(buildfile_dir)?

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+            buildfile_dir += args[last_index]
+
+            # add rest of dir name if not even amount of " or ' in dir name (dirs with spaces are wrapped)
+            while buildfile_dir.count('"') % 2 is not 0 or buildfile_dir.count('\'') % 2 is not 0:
+                last_index += 1
+                buildfile_dir += " " + args[last_index]
+
+        # remove ' or " and add / if missing
+        buildfile_dir = buildfile_dir.replace("'", "")
+        buildfile_dir = buildfile_dir.replace("\"", "")
```
salsolatragus: Use buildfile_dir.strip("'\\"") instead of these two lines. You don't need to do it if the condition in line 187 is false.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+            # dir is one index behind "-p" or "--project-dir" in args
+            last_index = 0
+            if "--project-dir" in args:
+                last_index = args.index("--project-dir") + 1
+            else:
+                last_index = args.index("-p") + 1
+            buildfile_dir += args[last_index]
+
+            # add rest of dir name if not even amount of " or ' in dir name (dirs with spaces are wrapped)
+            while buildfile_dir.count('"') % 2 is not 0 or buildfile_dir.count('\'') % 2 is not 0:
```
salsolatragus: You should be able to use shlex.split(command) instead of command.split() to save this manual effort here.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+                output = Shell.exec(command, cwd=project_dir, logger=logger)
                 dependencies = Compile._parse_ant_classpath(output)
                 Compile._copy_classpath(dependencies, dep_dir)
             elif command.startswith("gradle "):
-                Shell.exec(command,cwd=project_dir,logger=logger)
-                shutil.copy(os.path.dirname(__file__) + '/classpath.gradle', project_dir)
-                output = Shell.exec("gradle :printClasspath -b classpath.gradle", cwd=project_dir, logger=logger)
+                Shell.exec(command, cwd=project_dir, logger=logger)
+                buildfile_dir, classpath_cmd = Compile._parse_buildfile_dir(command)
+                shutil.copy(os.path.dirname(__file__) + '/classpath.gradle', project_dir + "/" + buildfile_dir)
```
salsolatragus: Use os.path.join() to build the paths.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+                output = Shell.exec(classpath_cmd, cwd=project_dir, logger=logger)
                 dependencies = Compile._parse_gradle_classpath(output)
                 Compile._copy_classpath(dependencies, dep_dir)
             else:
                 Shell.exec(command, cwd=project_dir, logger=logger)
 
     @staticmethod
+    def _parse_buildfile_dir(command):
```
salsolatragus: I expected this method to only parse the dir, not to also construct the classpath_command. Create the command string in the caller (like you do for Maven and Ant), then this method returns only a string, instead of a tuple.

Btw. the parse_... methods can all be private (i.e., start with two _), right?

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,24 +158,56 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+        # parsing for "-p" or "--project-dir" in gradle command
+        # and adjusting printClasspath command with it
+        # gradle :printClasspath -b buildfile_dir/classpath.gradle
+
+        args = command.split()
+        buildfile_dir = ""
+        if "-p" in args or "--project-dir" in args:
+            # dir is one index behind "-p" or "--project-dir" in args
+            last_index = 0
```
salsolatragus: Rename to dir_arg_index?

mubench.pipeline/tasks/implementations/compile.py
joschli: "gradle ... -p {} -b classpath.gradle".format(buildfile_dir) didnt work when i tried it

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,25 +159,46 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+                dependencies = Compile.__parse_gradle_classpath(output)
                 Compile._copy_classpath(dependencies, dep_dir)
             else:
                 Shell.exec(command, cwd=project_dir, logger=logger)
 
     @staticmethod
-    def _parse_maven_classpath(shell_output: str) -> List[str]:
+    def __parse_buildfile_dir(command):
+        # parsing for "-p" or "--project-dir" in gradle command
+        # and adjusting printClasspath command with it
```
salsolatragus: This no longer happens here. Remove the comment.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,25 +159,46 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+                output = Shell.exec(command, cwd=project_dir, logger=logger)
+                dependencies = Compile.__parse_gradle_classpath(output)
                 Compile._copy_classpath(dependencies, dep_dir)
             else:
                 Shell.exec(command, cwd=project_dir, logger=logger)
 
     @staticmethod
-    def _parse_maven_classpath(shell_output: str) -> List[str]:
+    def __parse_buildfile_dir(command):
+        # parsing for "-p" or "--project-dir" in gradle command
```
salsolatragus: I think this is clear from the `if "-p"/"--project-dir" in args` lines. Remove the comment.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,25 +159,46 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+        # and adjusting printClasspath command with it
+        # gradle :printClasspath -b buildfile_dir/classpath.gradle
+        args = shlex.split(command)
+        buildfile_dir = ""
+
+        # dir is one index behind "-p" or "--project-dir" in args
+        # remove ' or " and add / if missing
```
salsolatragus: Is removing quotes even necessary, or does shlex do it for us? In any case, I don't think we need a comment to explain this.

You no longer adding a / if missing. Remove the comment.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -158,25 +159,46 @@ def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
...
+        if "-p" in args:
```
salsolatragus: I like this new condition structure! Only one level of control flow is better.

Only I would eliminate a little duplication and pull the strip() after the condition (unless it's dropped entirely, because see above ;))

PR #14, opened 2017-05-17
Implemented the infrastructure for CLI versioning

M8is:
We can now implement subclasses of JavaInterface for new CLI versions. For example, JavaInterfaceV20170406 in [java_interface.py](https://github.com/stg-tud/MUBench/blob/131d39b0bf310fd826f216b14739729165290aa3/mubench.pipeline/data/java_interface.py) is the implementation of our current interface. 

The DetectorExecution now only passes a dictionary of detector arguments to the JavaInterface, which then independently handles the detector call. 

The correct JavaInterface must be provided by the detector. We currently just pass a string parameter, but this should be easily extendable. For example, we could read this string from a config file.

Also a short note about the current implementation of JavaInterfaceV20170406: This implementation now filters unknown detector arguments and outputs a fitting DEBUG message. This should guarantee that the actual calls are compatible with the java CLI version, without changing anything on the Java side.

Still to do: Implement a JavaInterface version which passes the arguments via a config file.


mubench.pipeline/data/detector.py
```diff
@@ -2,21 +2,20 @@
...
 from data.finding import Finding, SpecializedFinding
-
+from data.java_interface import JavaInterface
 
 class Detector:
     BASE_URL = "http://www.st.informatik.tu-darmstadt.de/artifacts/mubench/detectors"
-    CLI_VERSION = "v20170406"
 
-    def __init__(self, detectors_path: str, detector_id: str, java_options: List[str]):
+    def __init__(self, detectors_path: str, detector_id: str, java_options: List[str], cli_version: str = "20170406"):
```
salsolatragus: The version was "**v**20170406" before. Please fix this. We should switch this to use the version of the CLI maven artifact, we can do this in our meeting tomorrow.

mubench.pipeline/data/java_interface.py
```diff
@@ -0,0 +1,77 @@
...
+from logging import Logger
+from typing import Optional, List, Dict
+
+from data.project_version import ProjectVersion
+from utils.shell import Shell
+
+def _quote(value: str):
+    return "\"{}\"".format(value)
+
+class JavaInterface:
```
salsolatragus: Let's call this `RunnerInterface` instead.

mubench.pipeline/data/java_interface.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, jar_path: str, java_options: List[str]):
+        self.jar_path = jar_path
+        self.java_options = java_options
+
+    @staticmethod
+    def get(cli_version: str, jar_path: str, java_options: List[str]) -> 'JavaInterface':
+        interfaces = JavaInterface.__subclasses__()
+        matching_interfaces = [i for i in interfaces if i.version() == cli_version]
```
salsolatragus: We should assert that there is only one matching interface. Otherwise, we broke something!

mubench.pipeline/tests/data/test_detector_execution.py
```diff
@@ -176,20 +170,22 @@ def test_execute_per_misuse(self, get_dependencies_classpath_mock, shell_mock):
...
+                'training_src_path' : training_src_path,
+                'training_classpath' : training_classpath,
+                'target_src_path' : target_src_path,
+                'target_classpath' : target_classpath,
+                'dep_classpath' : dependencies_classpath + ":" + original_classpath
+            },
+            42, self.logger)
+
+
+@patch("data.java_interface.Shell")
```
salsolatragus: This `@patch` is unnecessary now, right?

mubench.pipeline/tests/data/test_java_interface.py
```diff
@@ -0,0 +1,75 @@
...
+        actual = JavaInterface.get(JavaInterfaceTestImpl.TEST_VERSION, "", dict())
+        assert_true(isinstance(actual, JavaInterfaceTestImpl))
+
+@patch("data.java_interface.Shell")
+class TestJavaInterfaceV20170406:
+    # noinspection PyAttributeOutsideInit
+    def setup(self):
+        self.uut = JavaInterfaceV20170406("-detector-", [])
+        self.logger = logging.getLogger("test")
+        self.version = create_version("-version")
```
salsolatragus: should be "-version-"

python_tests
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+realpath() {
+    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
+}
+
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
+
+docker run --rm -v "$SCRIPT_DIR":/mubench svamann/mubench nosetests -w mubench.pipeline/
```
salsolatragus: Please use mubench-ci here, move the file into `mubench.pipeline`, and call it `run_tests`. The `SCRIPT_DIR` should be fixable with an additional call to `dirname`.

mubench.pipeline/tests/data/test_detector_execution.py
```diff
@@ -93,46 +93,42 @@ def setup(self):
 
         self.logger = logging.getLogger("test")
 
-        self.__orig_shell_exec = Shell.exec
-        Shell.exec = MagicMock()
+        self.detector.java_interface.execute = MagicMock()
```
salsolatragus: Can we use the `JavaInterfaceTestImpl` here?

mubench.pipeline/data/java_interface.py
```diff
@@ -0,0 +1,77 @@
...
+
+    def _get_command(self, version: ProjectVersion, detector_arguments: Dict[str, str]) -> str:
+        detector_invocation = ["java"] + self.java_options + ["-jar", _quote(self.jar_path)]
+        detector_arguments = self._get_cli_args(detector_arguments)
+        command = detector_invocation + detector_arguments
+        command = " ".join(command)
+        return command
+
+    @staticmethod
+    def _get_cli_args(detector_arguments: Dict[str, str]) -> List[str]:
```
salsolatragus: I would prefer if this method would only escape the argument values and return a respective Dict. This way, joining the parts happens entirely in line 69. If you need stable order of keys for testing, please use an OrderedDict in the tests. Then the entries will stay in order of insertion.

mubench.pipeline/data/java_interface.py
```diff
@@ -0,0 +1,77 @@
...
+        "training_src_path",
+        "training_classpath",
+        "target_src_path",
+        "target_classpath",
+        "dep_classpath"
+    ]
+
+    @staticmethod
+    def version() -> str:
+        return "20170406"
```
salsolatragus: The version had a "v" prefix before.

mubench.pipeline/tests/data/stub_detector.py
```diff
@@ -3,4 +3,5 @@
 
 class StubDetector(Detector):
     def __init__(self):
-        super().__init__("-detectors-", "StubDetector", [], cli_version=RunnerInterfaceTestImpl.TEST_VERSION)
+        self._get_release_info = lambda *_: {'tag': 'latest', 'md5': None, 'cli_version': RunnerInterfaceTestImpl.TEST_VERSION}
```
salsolatragus: Do we need this hack to set the values? Since the releases file does not exist in the tests, the detector should use the defaults. This is fine for `tag` and `md5`, and we can set `self.cli_version = RunnerInterfaceTestImpl.TEST_VERSION` after calling the super constructor, can't we?

mubench.pipeline/data/detector.py
```diff
@@ -1,33 +1,38 @@
...
+        self.md5 = release_info.get("md5", None)
+        self.cli_version = release_info.get("cli_version", None)
 
-        if exists(md5_file):
-            with open(md5_file) as file:
-                md5 = file.read().strip()
-
-        return md5
+        self.jar_path = join(self.path, self.base_name + ".jar")
+        self.jar_url = "{}/{}/{}.jar".format(Detector.BASE_URL, self.cli_version, self.base_name)
```
salsolatragus: We need to include the tag in the URL. I'll move the files on the server accordingly.

mubench.pipeline/data/detector_execution.py
```diff
@@ -69,13 +65,11 @@ def __init__(self, run_mode: DetectorMode, detector: Detector, version: ProjectV
...
     def execute(self, compile_base_path: str, timeout: Optional[int], logger: Logger):
-        detector_invocation = ["java"] + self.detector.java_options + ["-jar", _quote(self.detector.jar_path)]
-        command = detector_invocation + self._get_detector_arguments(self.version.get_compile(compile_base_path))
-        command = " ".join(command)
+        detector_args = self._get_detector_arguments(self.version.get_compile(compile_base_path))
 
         start = time.time()
         try:
-            Shell.exec(command, logger=logger, timeout=timeout)
+            self.detector.runner_interface.execute(self.version, detector_args, timeout, logger)
```
salsolatragus: This crashed with a null reference, when the interface has been set to None in line 28 of detector.py. Can we, instead of None, use a `NoInterface` implementation that raises a more meaningful exception on execute?

mubench.pipeline/data/detector_execution.py
```diff
@@ -125,7 +119,7 @@ def _run_mode_detector_argument(self):
     def _get_findings_path(self):
         raise NotImplementedError
 
-    def _get_detector_arguments(self, project_compile: ProjectCompile):
+    def _get_command(self, project_compile: ProjectCompile):
```
salsolatragus: Why do we need this? I don't see an invocation. This has been moved to the interface, right?

mubench.pipeline/tests/data/stub_detector.py
M8is: The super constructor uses `cli_version` to get the runner interface, so we need to replace it before the call. I agree that this is not very readable though. We can definitely omit `tag` and `md5`, since they default to None anyway. As for `cli_version`, I am not sure what the best solution is here. We should probably discuss this in our meeting later.

mubench.pipeline/data/detector_execution.py
M8is: I was contemplating if the caller should check `if self.detector.runner_interface`. One way or another, we will probably have to do error handling explicitly when calling `execute`, since the interface doesn't have logging. That is why I avoided raising an exception in the interface.

mubench.pipeline/data/detector_execution.py
M8is: You're right, I deleted the implementations, but forgot the abstract declaration. Which reminds me that there is an API for abstract base classes in python called ABC [[https://pymotw.com/3/abc/]](https://pymotw.com/3/abc/).

mubench.pipeline/data/detector_execution.py
salsolatragus: Valid point. I think I would like the task to look like this

    try:
      self.detector.execute(...)  # internally delegates to the interface
    except NoCompatibleRunnerInterface as e:
      self.logger.error("Cannot run detector, because it has no compatible runner interface", e)
      return self.skip(...)
    except ...
      ...

instead of introducing a second level of error handling through an if. This we could realize by a `NoInterface` implementation of `RunnerInterface`, which raises the respective exception on execute ;)

salsolatragus:
I rebased the branch. Remember `git pull --rebase` and I'll keep my fingers of this code now ;)

mubench.pipeline/data/detector.py
```diff
@@ -11,24 +11,27 @@ class Detector:
...
 
-    def _get_release_info(self, releases_index_file_path: str) -> Dict[str, str]:
+        self.runner_interface = RunnerInterface.get(cli_version, self.jar_path, java_options)
+
+    def _get_release_info(self, releases_index_file_path: str, requested_release: Optional[str]) -> Dict[str, str]:
         if exists(releases_index_file_path):
             releases = read_yaml(releases_index_file_path)
+            if requested_release:
+                releases = [r for r in releases if r.get("tag", "latest") == requested_release]
             if len(releases) > 0:
```
M8is: We should probably add some error handling here. Sadly, we can't limit the choices in the CLI, as they depend on the detector.

detectors/DemoDetector/releases.yml
```diff
@@ -0,0 +1,2 @@
+- cli_version: v20170406
```
salsolatragus: This should be 0.0.8 now.

detectors/MuDetect/releases.yml
```diff
@@ -0,0 +1,2 @@
+- cli_version: v20170406
```
salsolatragus: Should be 0.0.8

detectors/MuMiner/releases.yml
```diff
@@ -0,0 +1,2 @@
+- cli_version: v20170406
```
salsolatragus: Should be 0.0.8

mubench.pipeline/data/detector.py
salsolatragus: Agreed. We should raise a ValueError to indicate that the tag did not exist. Theoretically, I think we could load the detectors on startup and use the releases info to validate the CLI arguments, but I wouldn't go this far at this point.

mubench.pipeline/data/runner_interface.py
```diff
@@ -0,0 +1,100 @@
...
+        "training_src_path",
+        "training_classpath",
+        "target_src_path",
+        "target_classpath",
+        "dep_classpath"
+    ]
+
+    @staticmethod
+    def version() -> str:
+        return "v20170406"
```
salsolatragus: Shouldn't this be 0.0.8 now as well?

mubench.pipeline/data/runner_interface.py
```diff
@@ -0,0 +1,100 @@
...
+    @staticmethod
+    def _get_cli_args(detector_arguments: Dict[str, str]) -> Dict[str, str]:
+        args = OrderedDict()
+        for key, value in detector_arguments.items():
+            args[key] = _quote(value)
+        return args
+
+class NoCompatibleRunnerInterface(Exception):
+    def __init__(self, version):
+        super().__init__("")
```
salsolatragus: I think this exception should have the message that we currently build in `detector_execution.py:89`. There it should only say `logger.error("Detector failed: %s", e)`.

mubench.pipeline/tests/data/stub_detector.py
salsolatragus: The question is, should the info be loaded by the constructor? Can we load it at a later point to avoid this race condition?

mubench.pipeline/tests/data/test_runner_interface.py
```diff
@@ -0,0 +1,96 @@
...
+from utils.shell import Shell, CommandFailedError
+from tests.test_utils.data_util import create_misuse, create_version, create_project
+from tests.test_utils.runner_interface_test_impl import RunnerInterfaceTestImpl
+
+class TestRunnerInterface:
+    def test_get_interface_version(self):
+        actual = RunnerInterface.get(RunnerInterfaceTestImpl.TEST_VERSION, "", dict())
+        assert_true(isinstance(actual, RunnerInterfaceTestImpl))
+
+    def test_get_interface_version_default_for_none(self):
```
salsolatragus: Why is this a relevant use case?

M8is: If a release file is missing the cli_version, the detector defaults it to None. We could also let the detector raise an exception instead, as None should probably not be a valid parameter value for the factory method. Should I change that?

salsolatragus: Yes, please!

mubench.pipeline/data/detector.py
```diff
@@ -1,33 +1,56 @@
...
-        md5 = None
+            if releases:
+                return releases[0]
+            else:
+                raise ValueError("No matching {} release for {}".format(self.id, requested_release))
 
-        if exists(md5_file):
-            with open(md5_file) as file:
-                md5 = file.read().strip()
+        return dict()
```
salsolatragus: Shouldn't this case also raise an exception, because detectors need to specify releases now?

mubench.pipeline/tests/data/test_detector.py
```diff
@@ -1,19 +1,67 @@
...
+    def test_raises_on_no_matching_release(self):
+        assert_raises(ValueError, self.setup_releases,
+                [{"md5": "-md5-", "tag": "-release-", "cli_version": "-version-"}],
+                "-unavailable_release-")
+
+    def setup_releases(self, releases, requested_release = None):
+        detector_id = "-detector-"
+        releases_index = join(self.temp_dir, detector_id, Detector.RELEASES_FILE)
+        write_yaml(releases, releases_index)
+        self.detector = Detector(self.temp_dir, detector_id, [], requested_release)
```
salsolatragus: I would prefer it, if the tests instantiated the Detector themselves, because the name `setup_releases` does not state that this happens here.

(With a little bit of luck you can quickly solve this by extracting a method with the first 3 statements and then inlining the original one. Curious to see whether PyCharm manages this correctly ^^)

PR #16, opened 2017-05-17
creating jars from classes dependencies

joschli:
creating jars from classes dependencies, changed from List to Set to avoid duplicates

### DIFF
--- a/mubench.pipeline/tasks/implementations/compile.py
+++ b/mubench.pipeline/tasks/implementations/compile.py
@@ -218,9 +218,6 @@ class Compile(ProjectVersionTask):
             else:
                 shutil.copy(dependency, dep_dir)
 
-
-
-
     @staticmethod
     def __copy_misuse_classes(classes_path, misuses, destination):
         remove_tree(destination)
--- a/mubench.pipeline/tasks/implementations/compile.py
+++ b/mubench.pipeline/tasks/implementations/compile.py
@@ -165,7 +165,7 @@ class Compile(ProjectVersionTask):
         return buildfile_dir
 
     @staticmethod
-    def __parse_maven_classpath(shell_output: str) -> List[str]:
+    def __parse_maven_classpath(shell_output: str) -> Set[str]:
         # shell_output looks like (possibly multiple times, once for each Maven module):
         # [INFO] Dependencies classpath:
         # /path/dep1.jar:/path/dep2.jar
@@ -179,7 +179,7 @@ class Compile(ProjectVersionTask):
         return classpath
 
     @staticmethod
-    def __parse_ant_classpath(shell_output: str) -> List[str]:
+    def __parse_ant_classpath(shell_output: str) -> Set[str]:
         # shell_output looks like:
         #   [javac] '-classpath'
         #   [javac] '/project/build:/path/dep1.jar:/path/dep2.jar'
@@ -192,7 +192,7 @@ class Compile(ProjectVersionTask):
         return classpath.split(":")
 
     @staticmethod
-    def __parse_gradle_classpath(shell_output: str) -> List[str]:
+    def __parse_gradle_classpath(shell_output: str) -> Set[str]:
         # shell_output looks like:
         # :printClasspath
         # /path/dependency1.jar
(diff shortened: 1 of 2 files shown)


mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -206,12 +206,19 @@ def __parse_gradle_classpath(shell_output: str) -> List[str]:
...
+    def _copy_classpath(dependencies: Set[str], dep_dir: str):
         remove_tree(dep_dir)
         makedirs(dep_dir, exist_ok=True)
         for dependency in dependencies:
             if not os.path.isdir(dependency):
                 shutil.copy(dependency, dep_dir)
+            elif dependency.endswith("classes"):
+                # dependency is a classes directory
+                # cutoff /target/classes to name jar
+                dep_name = os.path.basename(os.path.dirname(os.path.dirname(dependency)))
```
salsolatragus: Use path relative to checkout dir.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -206,12 +206,19 @@ def __parse_gradle_classpath(shell_output: str) -> List[str]:
         return lines[first_dependency_idx:first_empty_line_idx]
 
     @staticmethod
-    def _copy_classpath(dependencies: List[str], dep_dir: str):
+    def _copy_classpath(dependencies: Set[str], dep_dir: str):
         remove_tree(dep_dir)
         makedirs(dep_dir, exist_ok=True)
         for dependency in dependencies:
             if not os.path.isdir(dependency):
```
salsolatragus: Invert.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -206,17 +206,18 @@ def __parse_gradle_classpath(shell_output: str) -> Set[str]:
...
         for dependency in dependencies:
-            if not os.path.isdir(dependency):
-                shutil.copy(dependency, dep_dir)
-            elif dependency.endswith("classes"):
+            if os.path.isdir(dependency):
                 # dependency is a classes directory
-                # cutoff /target/classes to name jar
-                dep_name = os.path.basename(os.path.dirname(os.path.dirname(dependency)))
+                dep_name = os.path.relpath(dependency, checkout_base_path)
+                dep_name = dep_name.replace('/', '-')
```
salsolatragus: Please use `os.sep` instead of `/`.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -79,7 +79,7 @@ def process_project_version(self, project: Project, version: ProjectVersion):
                 logger.info("Compiling project...")
                 logger.debug("Copying patterns to source directory...")
                 self.__copy(version.patterns, sources_path)
-                self._compile(version.compile_commands, build_path, project_compile.dependencies_path)
+                self._compile(version.compile_commands, build_path, project_compile.dependencies_path, self.checkouts_base_path)
```
salsolatragus: Please use the `compiles_base_path` here. Although, currently, both are identical, we're handling files in the compiles directory here.

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -128,27 +128,27 @@ def __copy_additional_compile_sources(version: ProjectVersion, checkout_dir: str
             copy_tree(additional_sources, checkout_dir)
 
     @staticmethod
-    def _compile(commands: List[str], project_dir: str, dep_dir: str) -> None:
+    def _compile(commands: List[str], project_dir: str, dep_dir: str, checkout_base_path: str) -> None:
```
salsolatragus: please rename to compile_base_path (also in transitive methods)

mubench.pipeline/tests/tasks/implementations/test_compile.py
```diff
@@ -135,7 +135,7 @@ def test_passes_compile_commands(self):
 
         self.uut.process_project_version(self.project, self.version)
 
-        self.uut._compile.assert_called_with(["a", "b"], self.build_path, self.dep_path)
+        self.uut._compile.assert_called_with(["a", "b"], self.build_path, self.dep_path, self.checkout_base_path)
```
salsolatragus: use compiles path here as well

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -179,7 +179,7 @@ def __parse_maven_classpath(shell_output: str) -> List[str]:
         return classpath
 
     @staticmethod
-    def __parse_ant_classpath(shell_output: str) -> List[str]:
+    def __parse_ant_classpath(shell_output: str) -> Set[str]:
```
salsolatragus: What's the effect of this change? This is only a type int, isn't it? It doesn't change the type of the result, i.e., these methods still return lists with potential duplicates.

PR #19, opened 2017-05-31
Limit shown misuse size

joschli:


mubench.reviewsite/src/Model/Misuse.php
```diff
@@ -136,6 +136,12 @@ public function hasSufficientReviews()
         return $this->getReviewState() > ReviewState::NEEDS_REVIEW;
     }
 
+    public function isDistinctReviewState()
```
salsolatragus: I couldn't deduce what this function does from its name. I thought about renaming it to `hasConclusiveReviewState()`, but that's not it either, because its `true` on `NEEDS_REVIEW`. In lack of a good name that captures this intention, I would probably change this function to also return `false` on `NEEDS_REVIEW` and then additionally check for `!$misuse->hasSufficientReviews()` in the filtering. Or maybe you have a better suggestion?

mubench.reviewsite/src/RoutesHelper.php
```diff
@@ -19,24 +19,42 @@ class RoutesHelper
...
     }
 
     public function detector(Request $request, Response $response, array $args) {
+        $max_reviews = $request->getQueryParam("max_reviews", $this->max_reviews);
         $detector = $this->db->getOrCreateDetector($args['detector']);
-        $runs = $this->db->getRuns($detector, $args['exp']);
-        return $this->render($this, $request, $response, $args, 'detector.phtml', ['runs' => $runs]);
+        $runs = [];
+        foreach($this->db->getRuns($detector, $args['exp']) as $run){
+            $index = 0;
```
salsolatragus: You implemented this filtering logic twice, once here and once in DBConnection. Can we somehow pull this out into a function and reuse instead?

mubench.reviewsite/src/RoutesHelper.php
```diff
@@ -19,24 +19,42 @@ class RoutesHelper
...
 
-    public function __construct(DBConnection $db, PhpRenderer $renderer, Logger $logger, $upload_path, $site_base_url)
+    public function __construct(DBConnection $db, PhpRenderer $renderer, Logger $logger, $upload_path, $site_base_url, $max_reviews)
     {
         $this->db = $db;
         $this->renderer = $renderer;
         $this->logger = $logger;
         $this->site_base_url = $site_base_url;
         $this->upload_path = $upload_path;
+        $this->max_reviews = $max_reviews;
```
salsolatragus: Please rename to `$default_max_reviews`.

mubench.reviewsite/src/RoutesHelper.php
salsolatragus: This is not really an index, but the number of misuses with a conclusive review state. Please rename.

mubench.reviewsite/src/RoutesHelper.php
salsolatragus: `$default_ex2_max_reviews`

mubench.reviewsite/src/settings.php
```diff
@@ -18,13 +18,13 @@
...
-        'username'  => 'username',
-        'password'  => 'password',
+        'host'      => 'localhost:8889',
+        'database'  => 'mubench4',
+        'username'  => 'root',
+        'password'  => 'root',
         'charset'   => 'utf8',
         'collation' => 'utf8_unicode_ci',
-        'prefix'    => 'mubench_',
+        'prefix'    => 'mubench_fse17_',
```
salsolatragus: Please remove your local config again...

PR #20, opened 2017-06-06
stats rework

joschli:
stats are pulled out of the sent json ignoring potential_hits and detector, each detector got its own stats table and these are updated if a new stat column appears.

data from previous runs is deleted if a new run is pushed.

### DIFF
--- a/mubench.reviewsite/src/Controller/FindingsUploader.php
+++ b/mubench.reviewsite/src/Controller/FindingsUploader.php
@@ -38,7 +38,7 @@ class FindingsUploader
     {
         $propertyToColumnNameMapping = $this->getColumnNamesFromProperties($run);
         $propertyToColumnNameMapping = $this->removeDisruptiveStatsColumns($propertyToColumnNameMapping);
-        $this->createOrUpdateTable($detector->getStatsTableName(),$propertyToColumnNameMapping , array($this, 'createStatsTable'));
+        $this->createOrUpdateTable($detector->getStatsTableName(), $propertyToColumnNameMapping, array($this, 'createStatsTable'));
     }
 
     private function createOrUpdateFindingsTable(Detector $detector, $findings)
--- a/mubench.reviewsite/src/Controller/FindingsUploader.php
+++ b/mubench.reviewsite/src/Controller/FindingsUploader.php
@@ -25,27 +25,36 @@ class FindingsUploader
         $project = $run->{'project'};
         $version = $run->{'version'};
         $potential_hits = $run->{'potential_hits'};
-        $this->createOrUpdateTable($detector->getStatsTableName(), $run, array($this, 'createStatsTable'));
+
+        $this->createOrUpdateStatsTable($detector, $run);
         $this->deleteAndStoreStats($detector, $exp, $project, $version, $run);
         if ($potential_hits) {
-            $this->createOrUpdateTable($detector->getTableName(), $potential_hits, array($this, 'createFindingsTable'));
+            $this->createOrUpdateFindingsTable($detector, $potential_hits);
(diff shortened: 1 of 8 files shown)


mubench.reviewsite/tests/DatabaseTestCase.php
```diff
@@ -28,6 +28,7 @@ class DatabaseTestCase extends TestCase
         "result": "success",
         "runtime": 42.1,
         "number_of_findings": 23,
+        "-custom-stat-": "-stat-val-",
```
salsolatragus: Why do you need to touch the base test? What can we do to avoid changing the base test in case of an extension like this?

mubench.reviewsite/tests/StoreDetectorTest.php
```diff
@@ -21,6 +21,7 @@ function test_create_detector()
         "result": "failure",
         "runtime": 30.1,
         "number_of_findings": 0,
+        "-custom-stat-": "-stat-val-",
```
salsolatragus: Why do you need to touch this test? Storing detectors has nothing to do with storing run info, has it?

mubench.reviewsite/tests/StoreMetadataTest.php
```diff
@@ -29,7 +29,7 @@ function test_store_metadata()
             "result" => "success",
             "runtime" => "42.1",
             "number_of_findings" => "23",
-            "detector" => $detector->id,
+            "-custom-stat-" => "-stat-val-",
```
salsolatragus: Why do you have to touch this test? Storing meta data has nothing to do with storing run info, has it?

mubench.reviewsite/tests/StoreReviewTest.php
```diff
@@ -49,7 +49,7 @@ function setUp()
             "result" => "success",
             "runtime" => "42.1",
             "number_of_findings" => "23",
-            "detector" => $this->detector->id,
+            "-custom-stat-" => "-stat-val-",
```
salsolatragus: Why do you have to touch this test? Storing reviews has nothing to do with storing run info, has it?

mubench.reviewsite/src/Controller/FindingsUploader.php
```diff
@@ -24,45 +24,30 @@ public function processData($exp, $run)
...
-        if (count($propertyColumns) == 0) {
-            $this->createFindingsTable($detector);
-            $propertyColumns = ['exp', 'project', 'version', 'misuse', 'rank'];
+        $this->logger->info("Add columns to " . $table_name);
+        if(is_array($object)){
+            $propertyToColumnNameMapping = $this->getPropertyToColumnNameMapping($object);
+            $propertyToColumnNameMapping = $this->removeDisruptiveFindingsColumns($propertyToColumnNameMapping);
+        }else{
+            $propertyToColumnNameMapping = $this->getColumnNamesFromProperties($object);
+            $propertyToColumnNameMapping = $this->removeDisruptiveStatsColumns($propertyToColumnNameMapping);
```
joschli: any suggestions how we could improve this? I had to deduce what columns to remove from the given $object to remove only the columns in the specific context.

salsolatragus: How about you pass the `$propertyColumnMapping` as an argument into this function, instead of `$object`? Because on the call site you know what this anyways and you don't need the `$object` for anything else in here anyways, right?

By the way, I really hate that you keep calling variables `$object` or similar names. That's not a name, it's a type. You can do better.

joschli:
just found a bug with showing multiple projects after the rework. have to look into that before it can be merged.

PR #22, opened 2017-06-07
Soft requirements

M8is:
We check the requirements once using a `Check` task. 
Additionally, all requirements are "soft", meaning we don't force exit on missing requirements.

### DIFF
--- a/mubench.pipeline/task_runner.py
+++ b/mubench.pipeline/task_runner.py
@@ -5,7 +5,6 @@ from os.path import exists, join
 from typing import List
 
 from data.project import Project
-from requirements import are_satisfied
 from tasks.project_task import ProjectTask
 
 
@@ -22,8 +21,6 @@ class TaskRunner:
     def run(self) -> None:
         for task in self.tasks:
             logger = logging.getLogger()
-            logger.info("Checking requirements for %s...", task.name)
-            self.__check_requirements(task, logger)
             logger.info("Starting %s %s...", task.name, datetime.now().strftime("at %H:%M:%S"))
             task.start()
             task.black_list = self.black_list
@@ -39,17 +36,6 @@ class TaskRunner:
                             logger.info("Cannot proceed on %s; skipping for subsequent tasks.", project)
                         self.black_list.extend(response)
             task.end()
-
-    def check(self) -> None:
-        for task in self.tasks:
-            logger = logging.getLogger()
-            logger.info("Checking requirements for %s...", task.name)
-            self.__check_requirements(task, logger)
-
-    @staticmethod
-    def __check_requirements(task, logger):
-        if not are_satisfied(task.get_requirements(), logger):
-            logger.warning("Missing requirements for %s.", task.name)
 
     @staticmethod
     def _get_projects(data_path: str) -> List[Project]:
--- a/mubench.pipeline/task_runner.py
+++ b/mubench.pipeline/task_runner.py
@@ -49,8 +49,7 @@ class TaskRunner:
     @staticmethod
     def __check_requirements(task, logger):
         if not are_satisfied(task.get_requirements(), logger):
-            logger.error("Cannot run '%s'. Please ensure requirements.", task.name)
-            exit(1)
+            logger.warning("Missing requirements for %s.", task.name)
 
     @staticmethod
     def _get_projects(data_path: str) -> List[Project]:
--- a/mubench.pipeline/tasks/implementations/check.py
+++ b/mubench.pipeline/tasks/implementations/check.py
@@ -1,7 +1,7 @@
 import logging
 from typing import List
 
-from tasks.project_task import Requirement
+from requirements import Requirement
 from tasks.project_task import ProjectTask
 from utils.shell import Shell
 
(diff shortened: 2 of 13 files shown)


mubench.pipeline/tasks/implementations/check.py
```diff
@@ -0,0 +1,45 @@
...
+from tasks.project_task import ProjectTask
+from utils.shell import Shell
+
+
+class Check(ProjectTask):
+    def __init__(self):
+        super().__init__()
+
+    def start(self):
+        self._check_all_requirements()
```
salsolatragus: I would inline this, since it's only a single call.

mubench.pipeline/tasks/implementations/check.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def process_project(self, *_):
+        return self.ok()
+
+    def _check_all_requirements(self):
+        logger = logging.getLogger("requirements")
+        logger.info("Checking all requirements...")
```
salsolatragus: I'm not 100% sure, but the Task Runner should already output something like "Running Check...", so we should drop this reduce output. I would suggest we rename the task to `RequirementsCheck`, to make the output more expressive.

mubench.pipeline/tasks/implementations/check.py
```diff
@@ -0,0 +1,45 @@
...
+        import requirements
+        requirements = map(lambda requirement: requirement(), Requirement.__subclasses__())
+        if Check._are_satisfied(requirements, logger):
+            logger.info("All requirements satisfied. You're good to go.")
+        else:
+            logger.info("Unsatisfied requirements. Some MUBench tasks might work anyways, but to use the entire benchmark,"
```
salsolatragus: Make this a `logger.warning`

mubench.pipeline/tasks/implementations/check.py
```diff
@@ -0,0 +1,45 @@
...
+        return all_satisfied
+
+    @staticmethod
+    def _is_satisfied(requirement: Requirement, logger: logging.Logger) -> bool:
+        try:
+            requirement.check()
+            logger.debug("Requirement '%s' satisfied", requirement.description)
+            return True
+        except Exception as e:
+            logger.error("Requirement '%s' not satisfied: %s", requirement.description, e)
```
salsolatragus: Make this a `logger.warning`

mubench.pipeline/tests/tasks/implementations/test_check.py
```diff
@@ -0,0 +1,18 @@
...
+    def test_process_project(self):
+        assert_equals([], Check().process_project(create_project("-project-")))
+
+    def test_checks_requirements_on_start(self):
+        uut = Check()
+        uut._check_all_requirements = MagicMock()
+
+        uut.start()
+
+        uut._check_all_requirements.assert_called_once_with()
```
salsolatragus: This test basically just tests that the method you invoke is invoked. I don't see value in this. Can you mock the `Requirements.__subclasses__()` instead and return some test stub to control what happens?

mubench.pipeline/task_runner.py
```diff
@@ -40,18 +37,6 @@ def run(self) -> None:
...
 
-    def check(self) -> None:
-        for task in self.tasks:
-            logger = logging.getLogger()
-            logger.info("Checking requirements for %s...", task.name)
-            self.__check_requirements(task, logger)
-
-    @staticmethod
-    def __check_requirements(task, logger):
-        if not are_satisfied(task.get_requirements(), logger):
```
salsolatragus: The `Task.get_requirements()` method can now be deleted from `Task` and all its subclasses.

mubench.pipeline/tasks/implementations/check.py
M8is: I actually removed that output there, since I figured that its not the Task Runners job to know what the tasks will do. The output now looks like this:

> Starting benchmark...
> Starting check at 14:32:51...
>     Checking all requirements...
>     All requirements satisfied. You're good to go.

We can rename the task instead, if you would like to get rid of the extra line, but to me its a bit easier to read this way.

salsolatragus: I'm fine with removing the "Checking requirements" output from the runner. I agree that this is non of its business no more. I was referring to the "Starting check at..." line. The "check" part comes from the task name. By renaming the task it would become "Starting requirementscheck at...", which basically duplicates the subsequent line (modulo the unfortunate formatting, which I plan to improve by splitting the task name at camel case in the future, which would make this "Starting requirements check at ...").

mubench.pipeline/tasks/implementations/check.py
```diff
@@ -0,0 +1,46 @@
...
+            logger.info("All requirements satisfied. You're good to go.")
+        else:
+            logger.warning("Unsatisfied requirements. Some MUBench tasks might work anyways, but to use the entire benchmark,"
+                        " please ensure that your environment meets all requirements.")
+
+    def process_project(self, *_):
+        return self.ok()
+
+    def _get_requirements(self) -> List[Requirement]:
+        import requirements
```
salsolatragus: Why do you need the import here, if you have it in line 4?

M8is: Right. Has been removed.

PR #23, opened 2017-06-09
Interface changelogs and legacy checks

M8is:
We still need to discuss where the warning should occur. The problem is once again, that the don't have direct access to logging when creating the interface. Neither does the detector. I'm not really sure how we should approach this. 

The first layer where we use logging again is in `benchmark.py:105`, where we run `find_detector`. 
The next time we have logging in the interface is on execution, but I doubt we want to dump the changelogs on each runner execution. We could instead put the changelogs into files and only log a short warning with some hint to the changelogs.

### DIFF
--- a/mubench.pipeline/benchmark.py
+++ b/mubench.pipeline/benchmark.py
@@ -102,8 +102,9 @@ class Benchmark:
     def __get_detector(self):
         try:
             java_options = ['-' + option for option in self.config.java_options]
-            return find_detector(self.DETECTORS_PATH, self.config.detector, java_options, self.config.requested_release)
-        except ValueError as e:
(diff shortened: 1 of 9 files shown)


mubench.pipeline/benchmark.py
```diff
@@ -102,8 +102,9 @@ def __get_experiment(self):
     def __get_detector(self):
         try:
             java_options = ['-' + option for option in self.config.java_options]
-            return find_detector(self.DETECTORS_PATH, self.config.detector, java_options, self.config.requested_release)
-        except ValueError as e:
+            return find_detector(self.DETECTORS_PATH, self.config.detector, java_options,
+                    self.config.requested_release)
```
salsolatragus: This should either go back into the last line or be correctly indented.

PR #24, opened 2017-06-09
Add checks for system requirements (requires container rebuild)

M8is:
Added `psutil` for system information and `hurry.filesize` to pretty print memory size.
I wrote no unit tests, since we would have to mock the essential functionality anyway. Errors should be noticeable on Shippable.

Notes: 
- We exclude logical CPUs.
- Memory is given in bytes. With `hurry.filesize` the current value prints as `8G`.

### DIFF
--- a/mubench.pipeline/requirements.py
+++ b/mubench.pipeline/requirements.py
@@ -91,15 +91,25 @@ class CPUCountRequirement(Requirement):
         if _in_container():
             return self._get_container_cpu_count()
         else:
-            psutil = _try_import("psutil")
-            return psutil.cpu_count(logical=False)
+            return self._get_normal_cpu_count()
 
     def _get_container_cpu_count(self):
         try:
             with open("/sys/fs/cgroup/cpu/cpu.cfs_quota_us") as file:
-                return int(file.readline()) / 100000
+                quota = int(file.readline())
+
+            no_limit = quota < 0
+            if no_limit:
+                return self._get_normal_cpu_count()
+            else:
+                return quota / 100000
+
         except Exception as e:
             raise RuntimeError("Failed to check CPU count: {}".format(e))
+
+    def _get_normal_cpu_count(self):
+        psutil = _try_import("psutil")
+        return psutil.cpu_count(logical=False)
 
 
 class MemoryRequirement(Requirement):
--- a/mubench.pipeline/requirements.py
+++ b/mubench.pipeline/requirements.py
@@ -1,4 +1,3 @@
-from hurry.filesize import size
 from os.path import exists
 from utils.shell import Shell
 
@@ -92,7 +91,7 @@ class CPUCountRequirement(Requirement):
         if _in_container():
             return self._get_container_cpu_count()
         else:
-            import psutil
+            psutil = _try_import("psutil")
             return psutil.cpu_count(logical=False)
 
     def _get_container_cpu_count(self):
@@ -107,18 +106,18 @@ class MemoryRequirement(Requirement):
     MIN_MEMORY = 8589934592
 
     def __init__(self):
-        super().__init__("Memory >= {}".format(size(self.MIN_MEMORY)))
+        super().__init__("Memory >= {}".format(self._to_readable_size(self.MIN_MEMORY)))
 
     def check(self):
         memory = self._get_memory()
         if memory < self.MIN_MEMORY:
-            raise ValueError("Only {} of memory available.".format(size(memory)))
+            raise ValueError("Only {} of memory available.".format(self._to_readable_size(memory)))
 
     def _get_memory(self):
         if _in_container():
             return self._get_container_memory_limit()
         else:
-            import psutil
+            psutil = _try_import("psutil")
             return psutil.virtual_memory().total
 
     def _get_container_memory_limit(self):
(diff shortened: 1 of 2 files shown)


M8is:
I just noticed that the current solution does not work inside containers. Working on it.

M8is:
I'm not sure how well this works on other systems. Only tested it on Ubuntu 16.04 LTE.

mubench.pipeline/requirements.py
```diff
@@ -74,3 +77,56 @@ def __init__(self):
...
+    def check(self):
+        cpu_count = self._get_cpu_count()
+        if cpu_count < self.MIN_CPU_COUNT:
+            raise ValueError("Only {} CPUs available.".format(cpu_count))
+
+    def _get_cpu_count(self):
+        if _in_container():
+            return self._get_container_cpu_count()
+        else:
+            return psutil.cpu_count(logical=False)
```
salsolatragus: What happens to this check in the Docker environment?

M8is: I'm not sure what you mean.
_What happens when we run this check inside a docker container?_ Docker is so nice as to mount these system files for each container. Hence, we just read the information we need from the respective files.
_Why can we not just use psutil?_ Container do not have their "own" system environment. They just have a limited scope of the host system they are allowed to access. When psutil accesses the systems specs, it still sees the specs of the host system and not the constraints of the container.

Hopefully I covered you're question with that. Does this solution also work on mac? I'm not sure if the mounted folder structure is the same there.

salsolatragus: It does. In that case, however, why should we add the dependencies to the Docker image, if they are not used anyways? If you move the imports to happen in the `not _in_container()` branch, everything should work just fine, right?

M8is: You're right, I didn't even think about that.

salsolatragus: The magic of four eyes and (more importantly) two brains ;)

M8is: I just changed it. Still, I added another requirement to pretty print memory size, which is also used inside the container. However, `psutils` should not be needed in the container anymore.
Now that brings up another problem: we have different requirements. Now we would need two separate requirements files. I'm not sure if it wouldn't be more sensible to just build the container with `psutils` anyway.

salsolatragus: Agreed.

salsolatragus: I'll do this on Monday.

salsolatragus: Anyways, since these dependencies are actually requirements themselves now... how about we add a fallback in case people don't have them? For the CPUs I would just issue a warning stating that we couldn't determine the CPI count. For the memory, we could default to outputting the unformatted size.

M8is: Makes sense. Its more for convenience anyway.

M8is:
For clarity another two notes:

Docker defaults the value we use to check CPU constraints to -1. In that case we now fall back to checking the CPU of the host. 

In general, our CPU check is not directly checking the count. In theory, a container could be configured in a way which screws the value we're reading. Still, it should be good enough, considering the docker documentation suggests to only use the `--cpus` flag, which is consistent with our implementation.

mubench.pipeline/requirements.py
```diff
@@ -93,31 +91,43 @@ def _get_cpu_count(self):
...
         try:
             with open("/sys/fs/cgroup/cpu/cpu.cfs_quota_us") as file:
-                return int(file.readline()) / 100000
+                quota = int(file.readline())
+
+            no_limit = quota < 0
+            if no_limit:
+                return self._get_normal_cpu_count()
+            else:
+                return quota / 100000
```
salsolatragus: Could you add a comment as to why you divide by 100000? This is not obvious (to me).

mubench.pipeline/requirements.py
```diff
@@ -127,6 +137,20 @@ def _get_container_memory_limit(self):
         except Exception as e:
             raise RuntimeError("Failed to check memory: {}".format(e))
 
+    def _to_readable_size(self, memory: int) -> str:
```
salsolatragus: Please rename `memory` to `size_in_bytes`

mubench.pipeline/requirements.py
```diff
@@ -93,31 +91,43 @@ def _get_cpu_count(self):
...
+            raise ValueError("Only {} of memory available.".format(self._to_readable_size(memory)))
 
     def _get_memory(self):
         if _in_container():
             return self._get_container_memory_limit()
         else:
+            psutil = _try_import("psutil")
             return psutil.virtual_memory().total
 
     def _get_container_memory_limit(self):
```
salsolatragus: Is there an "unlimited" case for memory as well (like for CPUs)?

M8is: There isn't. Without a set limit, the value should be equal to the host system.

mubench.pipeline/requirements.py
```diff
@@ -98,11 +98,20 @@ def _get_container_cpu_count(self):
...
+                Using '--cpus=X', docker sets the cpu-period to 100000 and cpu-quota to 100000 * X.
+                The default cpu-quota is -1, hence we check the host sytem in that case.
+                Example: '--cpus=1.5' means '--cpu-period=100000 --cpu-quota=150000'.
+            """
             no_limit = quota < 0
             if no_limit:
                 return self._get_normal_cpu_count()
             else:
-                return quota / 100000
+                cpu_period = 100000
```
salsolatragus: Shouldn't we then retrieve the `cpu_period` also from the docker environment? It seems like a user could set it to a different value on the command line, which would mess up our logic here...

M8is: :+1:

PR #26, opened 2017-06-19
adding personal rank icons for positive reviews

joschli:


mubench.reviewsite/src/Model/Review.php
```diff
@@ -45,6 +45,21 @@ public function getHitViolationTypes($rank){
         return $this->getReviewFinding($rank)['violation_types'];
     }
 
+    public function hasPositiveHit()
```
salsolatragus: We have the two concepts "potential hit" and "hit". Please rename to `identifiesHit`.

mubench.reviewsite/src/Model/Review.php
```diff
@@ -45,6 +45,21 @@ public function getHitViolationTypes($rank){
...
+    {
+        return $this->getDecision() == Decision::YES;
+    }
+
+    public function getMinimalPositiveHitRank()
```
salsolatragus: Please rename to `getLowestHitRank`. Please also throw an exception, when there is no hit.

mubench.reviewsite/templates/detector.phtml
```diff
@@ -117,6 +117,40 @@ use MuBench\ReviewSite\Model\ReviewState;
...
                             (<a href="<?= $review_url ?>/resolution"><i class="fa fa-handshake-o" aria-hidden="true" title="Resolution"></i></a>)
                         <?php endif; ?>
+                        <?php if($user && $misuse->hasReviewed($user)): ?>
+                            <?php
+                                $user_review = $misuse->getReview($user);
+                                $reviews = $misuse->getReviews();
+                            ?>
+                            <?php if($review_state == ReviewState::NEEDS_REVIEW && $user_review->hasPositiveHit()) : ?>
+                                <i class="fa fa-thumbs-o-up" aria-hidden="true" style="color:green;" title="Agreement: Yes"></i>
+                                <span class="fa-stack fa-3x" style="font-size:0.7em">
```
salsolatragus: This span is repeated 4 times, with the only difference being the rank. Please extract into a include file and include 4 times.

mubench.reviewsite/templates/detector.phtml
```diff
@@ -117,6 +117,40 @@ use MuBench\ReviewSite\Model\ReviewState;
                             <i class="fa fa-thumbs-o-down" aria-hidden="true" style="color:darkred;" title="Resolved: No"></i>
                             (<a href="<?= $review_url ?>/resolution"><i class="fa fa-handshake-o" aria-hidden="true" title="Resolution"></i></a>)
                         <?php endif; ?>
+                        <?php if($user && $misuse->hasReviewed($user)): ?>
+                            <?php
+                                $user_review = $misuse->getReview($user);
+                                $reviews = $misuse->getReviews();
+                            ?>
+                            <?php if($review_state == ReviewState::NEEDS_REVIEW && $user_review->hasPositiveHit()) : ?>
```
salsolatragus: The condition `$user_review->hasPositiveHit()` should be true for all subsequent cases, i.e., if the user did not identify a hit, we never want to show no rank. I know that the subsequent cases, by checking for an agreement, implicitly also check that the user identified a hit. However, I'd prefer this check to wrap this entire thing, just like the `hasReviewed($user)`.

joschli: i thought we also show the rank of a resolved review, even if the personal review did not identify a hit? This would not be the case if we move it up.

salsolatragus: Fair enough, I missed this case.

PR #28, opened 2017-06-21
Build error cleanup

M8is:
We now use a `BuildCommand` which handles the command execution for a single command during the compile task.

For Maven and Gradle the error output is filtered. 
We still need a solution for Ant. I only found that errors are printed to `stderr`. We would have to extend `Shell` to access and "copy" that stream.


mubench.pipeline/data/build_command.py
```diff
@@ -0,0 +1,190 @@
...
+        self._copy_dependencies(output, project_dir, dep_dir, compile_base_path)
+
+    @staticmethod
+    def name() -> str:
+        raise NotImplementedError
+
+    def _get_command(self, args: List[str]) -> str:
+        return ' '.join([self._name] + self._extend_args(args))
+
+    def _extend_args(self, args: List[str]) -> List[str]:
```
salsolatragus: There's nothing restricting this method to only extension of the arguments, how about we name it `prepare_args`?

mubench.pipeline/data/build_command.py
```diff
@@ -0,0 +1,190 @@
...
+
+def _copy_classpath(dependencies: Set[str], dep_dir: str, compile_base_path: str):
+    remove_tree(dep_dir)
+    makedirs(dep_dir, exist_ok=True)
+    for dependency in dependencies:
+        if os.path.isdir(dependency):
+            # dependency is a classes directory
+            dep_name = os.path.relpath(dependency, compile_base_path)
+            dep_name = dep_name.replace(os.sep, '-')
+            _create_jar(dependency, os.path.join(dep_dir, dep_name + ".jar"))
```
salsolatragus: This class starts with a single `_` while the function declared below starts with two `__`!

mubench.pipeline/data/build_command.py
```diff
@@ -0,0 +1,190 @@
...
+from os.path import join, exists, dirname, splitext, relpath
+from typing import List, Set
+
+from utils.io import remove_tree, copy_tree
+from utils.shell import Shell, CommandFailedError
+
+
+class BuildCommand:
+    def __init__(self, name: str, args: List[str]):
+        self.logger = logging.getLogger("buildcommand")
```
salsolatragus: I would prefer to pass the logger into `execute` from the caller, because this is consistent with how `Shell.exec` works and it allows the caller to care about things like the correct nesting depth (currently, the output would not be indented, whereas before it was indented by two levels).

mubench.pipeline/tasks/implementations/compile.py
```diff
@@ -125,97 +126,8 @@ def __copy_additional_compile_sources(version: ProjectVersion, checkout_dir: str
...
-        makedirs(dep_dir, exist_ok=True)
-        for dependency in dependencies:
-            if os.path.isdir(dependency):
-                # dependency is a classes directory
-                dep_name = os.path.relpath(dependency, compile_base_path)
-                dep_name = dep_name.replace(os.sep, '-')
-                Compile.__create_jar(dependency, os.path.join(dep_dir, dep_name + ".jar"))
-            else:
-                shutil.copy(dependency, dep_dir)
+            BuildCommand.get(command).execute(project_dir, dep_dir, compile_base_path)
```
salsolatragus: I would prefer `create(command)` over `get(command)`.

mubench.pipeline/data/build_command.py
```diff
@@ -0,0 +1,190 @@
...
+    def _get_command(self, args: List[str]) -> str:
+        return ' '.join([self._name] + self._extend_args(args))
+
+    def _extend_args(self, args: List[str]) -> List[str]:
+        return args
+
+    def _get_errors(self, output: str) -> str:
+        return output
+
+    def _copy_dependencies(self, exec_output: str, project_dir: str, dep_dir: str, compile_base_path: str) -> None:
```
salsolatragus: All three implementations of this method end with the same two lines: retrieve dependency list and copy dependencies. It would be more consistent and a little less repetition if this would be `_get_dependencies`, which returns the dependency list and the copy would happen in the caller. This leads us to the problem that there are commands without dependencies, where the copy might fail (it wouldn't, currently, but I don't like to depend on this). I think the clean solution would be to have a subclass of `BuildCommand`, say, `BuildToolCommand`, which only adds the dependency handling and is the superclass of the three specialized commands we currently have... might also be overkill and we can just return an empty list by default and not invoke the copy if the list is empty. What do you think?

salsolatragus: Thinking more about it, it doesn't feel like copy dependencies should be the responsibility of the command at all... I think I would rather see it in the `CompileTask`, which means that the command returns the dependencies from execute.

mubench.pipeline/tests/data/test_build_command.py
```diff
@@ -0,0 +1,270 @@
...
+        try:
+            uut.execute("-p-", "-d-", "-bp-")
+        except CommandFailedError as e:
+            assert_equals("\n-output-", e.output)
+
+
+@patch("data.build_command.shutil.copy")
+@patch("data.build_command.Shell.exec")
+class TestMavenCommand:
+    def test_compile_with_maven(self, shell_mock, copy_mock):
```
salsolatragus: There should be some corresponding old tests that can be deleted now, right?

M8is: Right. I'll do a cleanup now that the commands are mostly set in place. I just wanted to make sure that everything still runs.

PR #49, opened 2017-07-03
violation types for multiselect are pulled from db table

joschli:


mubench.reviewsite/tests/StoreReviewTest.php
```diff
@@ -23,7 +23,7 @@ class StoreReviewTest extends DatabaseTestCase
             0 => [
                 'hit' => 'Yes',
                 'types' => [
-                    'missing/call'
+                    '1'
```
salsolatragus: How do you know `1` is the correct value here?

joschli: missing/call is the first to get inserted in our init_db.sql so it should get the id 1

salsolatragus: Hm... I'm not really happy with this, because a change at a theoretically independent location can break this test... however, I plan to cleanup this types thing soon anyways, so let's leave it like this for now.

Issue #56, opened 2017-07-17
First-line markers appear in other lists


salsolatragus:
There is a problem with the scoping of the line markers, e.g., here they appear in the violation-types list:
![screenshot 2017-07-17 15 59 24](https://user-images.githubusercontent.com/154050/28271539-1a601db0-6b09-11e7-9e0a-e1895d7a72fa.png)

PR #65, opened 2017-07-27
Fix #39: added margin and fixed offset calculation for line markers

### LINKED ISSUE #39: Location markers have an offset of 2 and don't show the rank
For example in http://mubench2.svamann.de/site/index.php/ex3/MuDetectXProject_170627_XLimitAndAllTypes/adempiere/1312/adempiere.2 all the markers are offset by two lines (e.g., 207 instead of 209). Furthermore, shouldn't the markers show the rank of the respective finding?

joschli:


joschli:
This fixes also Issue #56, because all code blocks have a unique id.

Issue #47, opened 2017-06-30
Import examples from Owolabi into dataset


salsolatragus:
http://fsl.cs.illinois.edu/spec-eval/

M8is:
I committed a demo of the current output. Its probably easier if to review that commit instead of the script itself.

M8is:
I pushed another demo of the current output. This contains all information I am able to assemble automatically. The remaining TODOs probably need manual work.

PR #68, opened 2017-08-09
Generated the Owolabi ASE'16 dataset

M8is:
I've generated the dataset, added the locations for misuses, and fixed the multi-line descriptions for #47. Sadly, there are less misuses than expected as there are tons of duplicated reports in the csv table. We're at **22 misuses**, 17 versions, and 14 projects in the new dataset.
Just for reference: the script I used is in `mubench.pipeline/load_spec_eval_dataset` and uses `<username> <token>` as args (optional, without the args the number of API requests GitHub accepts is very limited).


salsolatragus:
Too bad, but 22 additional misuses already enlarge our dataset by a third. That's quite something!

Can you explain why they have the duplicates in their tables? I mean, these make their findings appear much larger...

M8is:
I have no clue. For example, `https://github.com/cbeust/testng/pull/830` appears in 8 lines and maps to multiple different project names in the Manual Inspections table below.

salsolatragus:
Can this be forks?

M8is:
Possibly. But if so, they aren't linked anywhere. Some pull requests also fix the same violations in multiple places. This combined with forks may explain the duplication. 
This also means the current dataset probably contains misuses where the misuse is not only in one location. This does not fit in our current model, does it?

salsolatragus:
We currently model misuse instances, i.e., if the same problem occurs in multiple locations, the dataset should contain multiple (equal) misuses, each pointing to one of the locations.

M8is:
Then we can probably "split" some misuses. However, we should probably do that last, since the information that still needs to be added manually (e.g. characteristics) will be mostly the same anyway.

salsolatragus:
Fair enough. I'm not sure when I will get around to do that, since I won't include this extended dataset in the upcoming submission (2.5 weeks to go...). It may well be that this PR sleeps until after the deadline.

salsolatragus:
In `hoverruan-weiboclient4j` the misuse is not in the location designated by the PR, but in the constructor of `Cid`. I would argue that we should have one misuse should be located there, once, and not for every client. What do you think? The same is true for `jodatime.362` and the `Zone` constructor.

I through with generating patterns for all misuses. I generated multiple patterns for those misuses we need to duplicate for more locations. After duplication, every misuse should again have exactly one pattern.

salsolatragus:
Correct misuse locations:

* `jodatime.362` -> `Zone(StringTokenizer)` and `Zone(String, StringTokenizer)`
* `hoverruan-weiboclient4j.128` -> `Cid(String)` (the other cases correspondingly)

M8is:
We should probably change the patterns for `hoverruan-weiboclient4j` misuses, as their locations now point at the parseLong directly.

salsolatragus:
Yes, these patterns need adjustment. I'll take care of it.

I see two more projects I identified above need duplication of misuses, for which I don't see any duplication in your comments:

> The misuse in `ivantrendafilov-confucius` should be duplicated for all the additional locations fixed in the PR.

> The misuse `jodatime.269` occurs in multiple location, but all test code. Need to discuss whether we want to include this. If we do, we need to duplicate the misuse for all locations.

We didn't discuss the test-code case, if I remember correctly. I'm leaning towards: If others include test code, why shouldn't we? Strange behaviour in tests may be less critical, but they are probably as developer-time consuming as bugs in the production code.

Also, did you take care of this?:

> Project `altoros-ycsb` has no `version.yml` and `project.yml` files.

After we've resolve these points, I would do one last pass through the entire thing to ensure we didn't miss anything and then we're hopefully through 🙈

M8is:
> If others include test code, why shouldn't we? Strange behaviour in tests may be less critical, but they are probably as developer-time consuming as bugs in the production code.

I agree. Especially since these are misuses which could probably occur in production code as well.

M8is:
We should probably also run a dataset check on this. I'm 99% sure I forgot something.

salsolatragus:
Haha, fair enough. That's what we built the check for ;)
I'll run it once I've done my final pass.

salsolatragus:
I did my pass, but there's a couple more things to do:

* [ ] Rebase this on master
* [ ] Use the `$maven.default.classes` variable in the versions (and $maven.default.test-classes`, see #149)
* [ ] Run the dataset consistency check
* [ ] Run a compile

salsolatragus:
Thanks for the update. Running a compile now.

salsolatragus:
Compile errors in

* [ ] apache-gora version bb09d89
* [ ] logblock-logblock-2 version 40548aa
* [ ] mqtt version f438425

salsolatragus:
* logblock-loblock-2 fails because lombok fails to do its magic...
* tbyktu-ntru currently fails to compile class "SchönhageStrassen" in the Docker container. I assume this is due to the shell's file encoding...

PR #72, opened 2017-08-09
Resolve #60: system for adding tags to a misuse

### LINKED ISSUE #60: Introduce tagging per misuse
Introduce a tags table and a linking table between misuses/findings (ex, detector, project, version, misuse). Tags can be added/created/deleted from the review page.

joschli:
new tables specified in init_db.sql and routes for adding/deleting tags on misuses from the review page of a misuse.

### DIFF
--- a/mubench.reviewsite/init_db.sql
+++ b/mubench.reviewsite/init_db.sql
@@ -97,7 +97,7 @@ CREATE TABLE IF NOT EXISTS `misuse_tags` (
   `version` varchar(100) NOT NULL,
   `misuse` varchar(100) NOT NULL,
   `tag` int(11) NOT NULL,
-  KEY `lookup` (`exp`,`detector`,`project`,`version`,`finding`)
+  KEY `lookup` (`exp`,`detector`,`project`,`version`,`misuse`)
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 
 CREATE TABLE IF NOT EXISTS `tags` (
--- a/mubench.reviewsite/init_db.sql
+++ b/mubench.reviewsite/init_db.sql
@@ -92,12 +92,12 @@ CREATE TABLE IF NOT EXISTS `types` (
 
 CREATE TABLE IF NOT EXISTS `misuse_tags` (
   `exp` varchar(10) NOT NULL,
-  `detector` varchar(100) NOT NULL,
-  `project` varchar(100) NOT NULL,
-  `version` varchar(100) NOT NULL,
-  `misuse` varchar(100) NOT NULL,
+  `detector` varchar(10) NOT NULL,
+  `project` varchar(30) NOT NULL,
+  `version` varchar(30) NOT NULL,
+  `misuse` varchar(30) NOT NULL,
   `tag` int(11) NOT NULL,
-  KEY `lookup` (`exp`,`detector`,`project`,`version`,`misuse`)
+  PRIMARY KEY `lookup` (`exp`,`detector`,`project`,`version`,`misuse`,`tag`)
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 
 CREATE TABLE IF NOT EXISTS `tags` (
--- a/mubench.reviewsite/init_db.sql
+++ b/mubench.reviewsite/init_db.sql
@@ -95,7 +95,7 @@ CREATE TABLE IF NOT EXISTS `misuse_tags` (
   `detector` varchar(100) NOT NULL,
   `project` varchar(100) NOT NULL,
   `version` varchar(100) NOT NULL,
-  `finding` varchar(100) NOT NULL,
+  `misuse` varchar(100) NOT NULL,
   `tag` int(11) NOT NULL,
   KEY `lookup` (`exp`,`detector`,`project`,`version`,`finding`)
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
--- a/mubench.reviewsite/init_db.sql
+++ b/mubench.reviewsite/init_db.sql
@@ -90,6 +90,23 @@ CREATE TABLE IF NOT EXISTS `types` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 
+CREATE TABLE IF NOT EXISTS `misuse_tags` (
+  `exp` varchar(10) NOT NULL,
+  `detector` varchar(100) NOT NULL,
+  `project` varchar(100) NOT NULL,
+  `version` varchar(100) NOT NULL,
+  `finding` varchar(100) NOT NULL,
+  `tag` int(11) NOT NULL,
+  KEY `lookup` (`exp`,`detector`,`project`,`version`,`finding`)
+) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
+
+CREATE TABLE IF NOT EXISTS `tags` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(100) NOT NULL,
+  PRIMARY KEY (`id`),
+  UNIQUE KEY `name` (`name`)
+) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
+
 INSERT INTO types (name)
 VALUES
     ('missing/call'),
--- a/mubench.reviewsite/src/Controller/TagController.php
+++ b/mubench.reviewsite/src/Controller/TagController.php
@@ -1,0 +1,61 @@
+<?php
+
+namespace MuBench\ReviewSite\Controller;
+
+use Monolog\Logger;
+use MuBench\ReviewSite\DBConnection;
+
+class TagController
+{
+
+    private $db;
+    private $logger;
+
+    function __construct(DBConnection $db, Logger $logger)
+    {
+        $this->db = $db;
+        $this->logger = $logger;
+    }
+
+    public function saveTagForMisuse($tag)
+    {
+        $tag_id = $this->getOrCreateTag($tag['name'])['id'];
+        $experiment = $tag['exp'];
+        $detector_id = $tag['detector'];
+        $project = $tag['project'];
+        $version = $tag['version'];
+        $finding = $tag['finding'];
+        $this->logger->info("saving tag $tag_id for $experiment, $detector_id, " . $project . ", " . $version . ", " . $finding);
+        $misuse_tags = $this->db->getTagsForMisuse($experiment, $detector_id, $project, $version, $finding);
+        foreach($misuse_tags as $misuse_tag){
+            if($tag_id === $misuse_tag['id']){
+                return;
+            }
+        }
+        $this->db->table('misuse_tags')->insert(['exp' => $experiment, 'detector' => $detector_id,
+            'project' => $project,
+            'version' => $version, 'finding' => $finding, 'tag' => $tag_id]);
+    }
+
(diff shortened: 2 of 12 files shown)


joschli:
@salsolatragus this can be reviewed

mubench.reviewsite/init_db.sql
```diff
@@ -83,6 +83,23 @@ CREATE TABLE IF NOT EXISTS `types` (
...
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 
+CREATE TABLE IF NOT EXISTS `misuse_tags` (
+  `exp` varchar(10) NOT NULL,
+  `detector` varchar(100) NOT NULL,
+  `project` varchar(100) NOT NULL,
+  `version` varchar(100) NOT NULL,
+  `misuse` varchar(100) NOT NULL,
+  `tag` int(11) NOT NULL,
+  KEY `lookup` (`exp`,`detector`,`project`,`version`,`misuse`)
```
salsolatragus: You could include `tag` in the key and make this a PRIMARY KEY, since we can assign each tag only once. This will (for maaaaaany tags ;)) improve the efficiency of the delete query and prevent duplicated tags. Since this is a N:M linking table, its the usual procedure to have the key span the entire table (unless the link itself has additional properties).

mubench.reviewsite/init_db.sql
```diff
@@ -83,6 +83,23 @@ CREATE TABLE IF NOT EXISTS `types` (
   UNIQUE KEY `name` (`name`)
 ) ENGINE=MyISAM  DEFAULT CHARSET=latin1;
 
+CREATE TABLE IF NOT EXISTS `misuse_tags` (
+  `exp` varchar(10) NOT NULL,
+  `detector` varchar(100) NOT NULL,
+  `project` varchar(100) NOT NULL,
+  `version` varchar(100) NOT NULL,
+  `misuse` varchar(100) NOT NULL,
```
salsolatragus: Note that I reduced the max length of detector to 10 (it should be an int anyways...) and of project, version, and misuse to 30, because the database on my production server complained about "key too large" otherwise. I thing that these sizes are sufficient, so please use them here, too.

mubench.reviewsite/src/Controller/TagController.php
```diff
@@ -0,0 +1,74 @@
...
+        $detector = $misuse['detector'];
+        $project = $misuse['project'];
+        $version = $misuse['version'];
+        $misuse_id = $misuse['misuse'];
+        $this->logger->info("saving tag $tag_id for $experiment, $detector, " .
+            $project .
+            ", " .
+            $version .
+            ", " .
+            $misuse_id);
```
salsolatragus: Why do some inline variables and some explicit concatenations? Inline all. Make it a one-liner.

mubench.reviewsite/src/Controller/TagController.php
```diff
@@ -0,0 +1,74 @@
...
+            ->delete();
+    }
+
+    private function getOrCreateTag($name)
+    {
+        $tag = $this->db->table('tags')->where('name', $name)->get();
+        if (!$tag) {
+            $this->saveNewTag($name);
+        }
+        return $this->db->table('tags')->where('name', $name)->get()[0];
```
salsolatragus: Isn't there a `getFirst()` or something?

mubench.reviewsite/templates/review.phtml
```diff
@@ -335,6 +397,18 @@ use MuBench\ReviewSite\Model\Experiment;
                     }
                     b.type = "hidden";
                 }
+                function onInput() {
+                    console.log("Hello");
+                    var val = document.getElementById("tag-input").value;
+                    var opts = document.getElementById("tags").childNodes;
```
salsolatragus: I assume these are the assigned tags? Please name them so.

mubench.reviewsite/templates/review.phtml
```diff
@@ -335,6 +397,18 @@ use MuBench\ReviewSite\Model\Experiment;
                     }
                     b.type = "hidden";
                 }
+                function onInput() {
+                    console.log("Hello");
+                    var val = document.getElementById("tag-input").value;
```
salsolatragus: `val` -> `tag_name_input`?

mubench.reviewsite/templates/review.phtml
```diff
@@ -335,6 +397,18 @@ use MuBench\ReviewSite\Model\Experiment;
...
+                    var opts = document.getElementById("tags").childNodes;
+                    var form = document.getElementById("tag-form");
+                    for (var i = 0; i < opts.length; i++) {
```
salsolatragus: It's not clear to me what this loop does. Can you make it more clear or add an explanatory comment?

mubench.reviewsite/tests/StoreMetadataTest.php
```diff
@@ -42,7 +42,8 @@ function test_store_metadata()
                     'method' => '-method-',
                     'diff_url' => '-diff-',
                     'snippets' => [0 => ['line' => '273', 'snippet' => '-code-']],
-                    'patterns' => [0 => ['name' => '-p-id-', 'code' => '-pattern-code-','line' => '1']]
+                    'patterns' => [0 => ['name' => '-p-id-', 'code' => '-pattern-code-','line' => '1']],
+                        'tags' => []
```
salsolatragus: Indent too large.

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,128 @@
+<?php
+
+require_once "DatabaseTestCase.php";
+
+use MuBench\ReviewSite\Controller\TagController;
+
+class TagTest extends DatabaseTestCase
```
salsolatragus: You're missing a test for adding a tag that did not previously exist.

mubench.reviewsite/templates/review.phtml
```diff
@@ -335,6 +397,18 @@ use MuBench\ReviewSite\Model\Experiment;
                     }
                     b.type = "hidden";
                 }
+                function onInput() {
+                    var tag_name_input = document.getElementById("tag-input").value;
+                    var suggested_tags = document.getElementById("tags").childNodes;
+                    var form = document.getElementById("tag-form");
+                    // loops over tags in the autocompletion list to submit the form when one is clicked
```
salsolatragus: Since this is triggered when the input changes, can I ever insert a tag `blafoo` when a tag `bla` exists? To my understanding, this listener will submit the form once I type `bla`. Isn't there a way to register an onClick listener on the datalist options or something?

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $misuse_tag1 = [
+            "tag" => 'test1',
+            "exp" => 1,
+            "detector" => 1,
+            "project" => "-project-",
+            "version" => "-version-",
+            "misuse" => "-misuse-"
+        ];
+        $misuse_tag2 = [
```
salsolatragus: Why is it important to add a second tag here? If it is, we should have a test for adding a single tag and a second for adding two tags, because adding a single tag is definitely a use case.

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,158 @@
...
```
salsolatragus: Same here, why is it important to delete two tags?

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,158 @@
...
```
salsolatragus: Adding a second tag is definitely not part of this test case. This has been tested above. Please remove this.

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,158 @@
...
+    {
+        $misuse_tag1 = [
+            "tag" => 'unknown_tag',
+            "exp" => 1,
+            "detector" => 1,
+            "project" => "-project-",
+            "version" => "-version-",
+            "misuse" => "-misuse-"
+        ];
+        $misuse_tag2 = [
```
salsolatragus: Also adding a known tag is not part of this test case. Please remove it.

mubench.reviewsite/tests/TagsTest.php
```diff
@@ -0,0 +1,158 @@
...
+class TagTest extends DatabaseTestCase
+{
+
+    private $tagController;
+
+    function setUp()
+    {
+        parent::setUp();
+        $this->tagController = new TagController($this->db, $this->logger);
+        $this->db->table('tags')->insert(['name' => 'test1']);
```
salsolatragus: Please move the insertion into the test cases (where it is needed). There's actually only one test that needs all three tags. The others need at most one or even none. Use `setUp` to reduce duplication in tests, not to introduce additional duplication.

mubench.reviewsite/templates/review.phtml
```diff
@@ -76,6 +94,50 @@ use MuBench\ReviewSite\Model\Experiment;
...
+                        <? if($user): ?>
+                        <div style="display:inline-block;height:100%;padding:4px;">
+                            <form id="tag-form" action="<?= $api_url_prefix ?>upload/tag" method="post" target="">
+                                <input type="hidden" name="exp" value="<?= htmlspecialchars($experiment->getId()); ?>">
+                                <input type="hidden" name="detector" value="<?= htmlspecialchars($detector->id); ?>">
+                                <input type="hidden" name="project" value="<?= htmlspecialchars($misuse->getProject()); ?>">
+                                <input type="hidden" name="version" value="<?= htmlspecialchars($misuse->getVersion()); ?>">
+                                <input type="hidden" name="misuse" value="<?= htmlspecialchars($misuse->id); ?>">
+                                <input type="hidden" name="path" value="<?= htmlspecialchars($path); ?>">
+                                <input id="tag-input" class="tag-input" onInput='onInput()' list="tags" name="tag" placeholder="add new tag" autocomplete="off">
```
salsolatragus: Please set the max length of the input to 100, to ensure it fits into the database.

mubench.reviewsite/templates/review.phtml
joschli: you are right, i did not think about that. the events dont work on the datalist tag.

salsolatragus: It seems there's no easy way around this... let's keep it this way for now.

salsolatragus: Maybe add this link: https://stackoverflow.com/questions/30022728/perform-action-when-clicking-html5-datalist-option as a reference for why we do it this way. Maybe there eventually a better way, which will hopefully be documented in the thread ;)

salsolatragus:
I think I had a misconception of how the completion list works. Am I correct that, without the handler, I can type and auto complete and the form will be submitted on Enter (does the Enter from selecting via the keyboard trigger a submit here?) and, with the handler, I can also click on an entry from the list, which with auto complete and submit? If this is the case, then I would actually prefer the variant without the handler. Somehow, I thought that without the handler we couldn't use autocomplete... sorry for the confusion.

joschli:
Without the Handler you have to press Enter to submit, but the autocomplete does still work. With the Handler you can also click an entry in the list and it would autocomplete and submit it.

salsolatragus:
Thanks for the clarification. Without the handler it is, then.

joschli:
@salsolatragus I removed the handler again, this should be ready to merge.

Issue #67, opened 2017-07-27
Dataset consistency check


salsolatragus:
Check if

*Project* has
* name
* repository (type, url)
* No version and misuse with same name

*Version* has
* revision
* at least one (existent!) misuse
* build (classes, src, commands (at least one))

*Misuse* has
* link from at least one version
* location. If a checkout exists, verify the misuse location (extract snippet), otherwise output warning.
* api
* characteristics (at least one)
* description (non-empty)
* fix (commit, revision)
* internal
* report
* source (name, url)

*Datasets.yml*:
* Ensure all entries exist

PR #91, opened 2017-08-21
Implement a dataset consistency check

M8is:
Implements #67

Misuse filter arguments may screw with some of the logic. We probably need to check for the white and black list in some cases, or omit the whole filter arguments feature and only allow a global check. I found it to be quite useful in some cases though. (Mainly to skip known issues.)

Logging seems a bit weird right now. I'm not sure how to improve it.

### DIFF
--- a/mubench.pipeline/tasks/implementations/dataset_check.py
+++ b/mubench.pipeline/tasks/implementations/dataset_check.py
@@ -1,7 +1,8 @@
 import logging
 
 from typing import Dict, List, Optional, Any
-from os.path import join
+from os import listdir
+from os.path import join, isdir
 
 from data.misuse import Misuse
 from data.project import Project
@@ -11,11 +12,16 @@ from tasks.project_version_misuse_task import ProjectVersionMisuseTask
 
 
 class DatasetCheck(ProjectVersionMisuseTask):
-    def __init__(self, datasets: Dict[str, List[str]], checkout_base_path: str):
+    def __init__(self, datasets: Dict[str, List[str]], checkout_base_path: str, data_base_path: str):
         super().__init__()
         self.logger = logging.getLogger("datasetcheck")
         self.datasets = datasets
         self.checkout_base_path = checkout_base_path
+        self.data_base_path = data_base_path
+        self.misuses_not_listed_in_any_version = []
+
+    def start(self):
+        self.misuses_not_listed_in_any_version = self._get_all_misuses(self.data_base_path)
 
     def process_project(self, project: Project):
         self._new_known_datasets_entry(project.id)
@@ -30,6 +36,7 @@ class DatasetCheck(ProjectVersionMisuseTask):
 
     def process_project_version_misuse(self, project: Project, version: ProjectVersion, misuse: Misuse):
         self._new_known_datasets_entry(misuse.id)
+        self._misuse_listed_in_version(misuse.id)
         self._check_required_keys_in_misuse_yaml(project, version, misuse)
         self._check_version_misuse_id_conflict(project.id, version.version_id, misuse.misuse_id)
         self._check_misuse_location_exists(project, version, misuse)
@@ -37,6 +44,7 @@ class DatasetCheck(ProjectVersionMisuseTask):
 
     def end(self):
         self._report_unknown_dataset_entries()
+        self._report_misuses_not_listed_in_any_version()
 
     def _check_required_keys_in_project_yaml(self, project: Project):
         yaml_path = "{}/project.yml".format(project.id)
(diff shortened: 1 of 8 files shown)


mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+        self._check_required_keys_in_version_yaml(project, version)
+        self._check_misuses_listed_in_version_exist(version)
+        return super().process_project_version(project, version)
+
+    def process_project_version_misuse(self, project: Project, version: ProjectVersion, misuse: Misuse):
+        self.logger = logging.getLogger("datasetcheck.project.version.misuse")
+        self._new_known_datasets_entry(misuse.id)
+        self._misuse_listed_in_version(misuse.id)
+        self._check_required_keys_in_misuse_yaml(project, version, misuse)
+        self._check_version_misuse_id_conflict(project.id, version.version_id, misuse.misuse_id)
```
salsolatragus: Since we create the ids as `project.version` or `project.misuse`, we may also have conflicts between a misuse and version that the misuse is not linked to.

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+
+    def _check_version_misuse_id_conflict(self, project_id: str, version_id: str, misuse_id: str):
+        if version_id == misuse_id:
+            self._version_misuse_conflict(project_id, version_id)
+
+    def _check_misuse_location_exists(self, project: Project, version: ProjectVersion, misuse: Misuse):
+        if "location" in misuse._yaml:
+            location = misuse.location
+            if location.file and location.method:
+
```
salsolatragus: remove this empty line

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+                checkout = version.get_checkout(self.checkout_base_path)
+                if not checkout.exists():
+                    self.logger.info(
+                            'Cannot check location for "{}": requires checkout of "{}".'.format(
+                                misuse.id, version.id))
+                    return
```
salsolatragus: Convert this into an if/else please.

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+    def _report_unknown_dataset_entries(self):
+        for dataset, entries in self.datasets.items():
+            for entry in entries:
+                self._unknown_dataset_entry(dataset, entry)
+
+    def _report_misuses_not_listed_in_any_version(self):
+        for misuse in self.misuses_not_listed_in_any_version:
+            self._misuse_not_listed(misuse)
+
+    def _missing_key(self, tag: str, file_path: str):
```
salsolatragus: Please rename all the method that log something to `_report_X` to make their function more clear.

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+                    return
+
+                source_base_path = join(checkout.checkout_dir, version.source_dir)
+                if not self._location_exists(source_base_path, location.file, location.method):
+                    self._cannot_find_location(str(location), "{}/misuses/{}/misuse.yml".format(project.id, misuse.misuse_id))
+
+    def _location_exists(self, source_base_path, file_, method) -> bool:
+        return get_snippets(source_base_path, file_, method)
+
+    def _new_known_datasets_entry(self, id_: str):
```
salsolatragus: `register` or `add` instead of `new`? `legal` instead of `known`? Not 100% happy with my suggestions, but also not with the current name.

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -0,0 +1,213 @@
...
+        for dataset, entries in self.datasets.items():
+            if id_ in entries:
+                entries.remove(id_)
+
+    def _misuse_listed_in_version(self, misuse: str):
```
salsolatragus: `register_misuse_is_linked_from_a_version`? Ensure this name is consistent with the method above.

salsolatragus:
Regarding the problems with the `--skip`/`--only` flags, I see two alternative solutions:

1. Disable any such checks if whitelist or blacklist are non-empty (and issue a respective warning).
2. Always run the full check, but disable the output depending on the filtering.

What do you think?

M8is:
Ready for review. 

I've added a warning when using filters. My reasoning is that this allows us to check the consistency of a single dataset. 
The only invalid warnings for a dataset should be at the very end, when existing keys from datasets and unused misuses are checked. At that point, IDs that are not in the dataset anyway can be ignored. 
All other warnings are still relevant in the context of that dataset (e.g. ID conflicts, missing keys, versions without misuses, ...).
For example, if you only care about the consistency of the icse16ex1 dataset, you can still use `mubench dataset-check --dataset icse16ex1`.

A workaround could be to remove IDs in the filter lists from our set of IDs to check. That should eliminate the invalid warnings. However, that's just a quick thought. It may lead to other problems...

PR #95, opened 2017-08-24
Resolve #88: added tags column to detector page

### LINKED ISSUE #88: Display tags on detector overview after review is complete
Last column.

joschli:
tags are only showing up if the misuse was has sufficient reviews.


mubench.reviewsite/templates/detector.phtml
```diff
@@ -231,6 +232,15 @@ $markdown_parser = new Parsedown();
                 <td class="<?= $style_classes ?>" style="max-width: 300px;">
                     <?= $misuse->hasReviewed($user) ? $misuse->getReview($user)->getComment() : "" ?>
                 </td>
+                <td class="<?= $style_classes ?>">
+                    <? if($misuse->hasSufficientReviews()) : ?>
```
salsolatragus: There needs to be an agreement/resolution. See my reply to your respective question in #88.

joschli:
Tags are now only shown if we have a conclusive reviewstate.

Issue #46, opened 2017-06-30
File access conflicts on Windows


salsolatragus:
Apparently, we sometimes get file-access conflicts, especially with svn checkouts on Windows, when files in the mounted `/mubench/` directory are accessed from within Docker and from the OS at the same time. I think the best solution would be to use a directory within docker for the processing and copy over the results into the mounted directory. Question is: Can we do this in a way that works both within and outside the Docker container?

joschli:
We would create the temp dir outside the mounted directory. This should also work outside the container or did i miss something?

salsolatragus:
Yes, only we need a way to determine that temp directory such that its location conforms to the environment standard, whatever that may be ;)

PR #97, opened 2017-08-28
Temp dir option

M8is:
Implementation for #46.
Compile now has the option to use a temporary directory. On successful completion, the directory is moved to the persistent standard directory. 
The new option is enabled by the `--windows-fix` flag, which is set in [mubench.bat](https://github.com/stg-tud/MUBench/blob/temp-dir-option/mubench.bat). Currently, this flag is hidden, i.e. not shown in the `mubench -h` text.

Note that checkouts still only use MUBench subdirectories. I am working on implementing that as well, but it is more complicated due to the local child repositories we create.

### DIFF
--- a/mubench.pipeline/utils/command_line_util.py
+++ b/mubench.pipeline/utils/command_line_util.py
@@ -1,3 +1,4 @@
+import argparse
 from argparse import ArgumentParser, HelpFormatter, ArgumentTypeError
 from operator import attrgetter
 
@@ -45,6 +46,9 @@ def get_command_line_parser(available_detectors: List[str], available_scripts: L
     subparsers = parser.add_subparsers(
         help="MUBench provides several tasks. Run `mubench <task> -h` for details.",
         dest='task')
+
+    parser.add_argument('--windows-fix', dest='windows_fix', default=get_default('windows-fix', False),
+                        help=argparse.SUPPRESS, action='store_true')
 
     __add_check_subprocess(subparsers)
     __add_info_subprocess(available_datasets, subparsers)
--- a/mubench.bat
+++ b/mubench.bat
@@ -11,5 +11,5 @@ IF "%~1" == "configure" (
     echo "Choose one of {review-site}"
   )
 ) ELSE (
-  docker run --rm -v "%SCRIPT_DIR%":/mubench svamann/mubench python ./mubench.pipeline/benchmark.py %*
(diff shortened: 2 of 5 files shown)


mubench.pipeline/utils/command_line_util.py
```diff
@@ -46,6 +47,9 @@ def get_command_line_parser(available_detectors: List[str], available_scripts: L
         help="MUBench provides several tasks. Run `mubench <task> -h` for details.",
         dest='task')
 
+    parser.add_argument('--windows-fix', dest='windows_fix', default=get_default('windows-fix', False),
```
salsolatragus: I'd prefer to call this `--use-tmp-wrkdir`, more descriptive, less blame ;)

M8is:
[Checkout (task)](https://github.com/stg-tud/MUBench/blob/temp-dir-option/mubench.pipeline/tasks/implementations/checkout.py) cannot just pass another directory to [ProjectCheckout](https://github.com/stg-tud/MUBench/blob/temp-dir-option/mubench.pipeline/data/project_checkout.py), since exist and delete require the actual path. 
I've tested to implement the option in ProjectCheckout, but the create procedures vary a lot. I'm not sure, if I missed some subclass where the current implementation doesn't work.

I'll try to get this stable today and push it later.

salsolatragus:
Could you create two `ProjectCheckout` instances the the checkout task: one on the MUBench path to check exists and delete and a second on the tmp dir, to do a new checkout, which is then copied over?

M8is:
I'm not sure if this solution collides with some other features, but it worked in my test runs so far.

I just noticed: `--force-checkout` leads to new checkouts for each version since the base checkout is deleted for each child, doesn't it?

mubench.pipeline/tasks/implementations/checkout.py
```diff
@@ -31,7 +34,15 @@ def process_project_version(self, project: Project, version: ProjectVersion) ->
                 logger.debug("Already checked out %s.", version)
             else:
                 logger.info("Fetching %s from %s...", version, checkout)
-                checkout.create()
+                if self.use_temp_dir:
+                    temp_dir = mkdtemp(prefix="mubench-checkout_")
+                    temp_checkout = version.get_checkout(temp_dir)
+                    temp_checkout.create()
```
salsolatragus: For the repository checkout this means that we clone from the remote for every version, since the base checkout doesn't exist in the temp directory, right? Not sure we can do anything about that... so I guess I could live with it, seeing that this is only a Windows workaround and checkout doesn't happen frequently...

Issue #74, opened 2017-08-09
Handle invalid file location for detector misuse


govindsinghnegi:
This exception has been observed only with Closure project and the other projects (Itext, Lucene etc.) are working fine. For some reason, the existing Findbugs code to return the Class source path using  "BugInstance.getPrimarySourceLineAnnotation().getSourcePath();" returns an incorrect path. 
A sample output is: "com/google/javascript/jscomp/**_\<Unknown\>_**"
whereas the correct path should have been: "com/google/javascript/jscomp/**_Class-name.java_**"
As you can see that the "class-name.java" part is missing. This leads to a "FileNotFound" exception being thrown while publishing the result to the review site. Logs and yml files are attached.

[logs_and_yml.zip](https://github.com/stg-tud/MUBench/files/1212027/logs_and_yml.zip)

salsolatragus:
In general, a detector returning invalid file paths is a problem that needs fixing on the detector side. However, we should add some more graceful handling of this situation, i.e., not crash and emit a helpful error message.

PR #124, opened 2017-09-07
Handle invalid locations

M8is:
Fix for #74.

We now catch that case and print a short warning.

### DIFF
--- a/mubench.pipeline/tasks/implementations/publish_findings_task.py
+++ b/mubench.pipeline/tasks/implementations/publish_findings_task.py
@@ -102,8 +102,9 @@ class PublishFindingsTask(ProjectVersionTask):
 
         try:
             snippets = finding.get_snippets(version_compile.original_sources_path)
-        except:
-            logger.warning("No snippets added: invalid file location.")
+        except SnippetUnavailableException as e:
+            logger.warning("No snippets added for %s", e.file)
+            logger.debug("Failed to find snippets: %s", e.exception)
 
         markdown_dict["target_snippets"] = [snippet.__dict__ for snippet in snippets]
         return markdown_dict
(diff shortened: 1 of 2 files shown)


mubench.pipeline/tasks/implementations/publish_findings_task.py
```diff
@@ -97,9 +97,14 @@ def __slice_by_max_files_per_post(self, potential_hits: List[SpecializedFinding]
...
         yield potential_hits_slice
 
-    def _prepare_post(self, finding: SpecializedFinding, version_compile) -> Dict[str, str]:
+    def _prepare_post(self, finding: SpecializedFinding, version_compile, logger) -> Dict[str, str]:
         markdown_dict = self._to_markdown_dict(finding)
-        snippets = finding.get_snippets(version_compile.original_sources_path)
+
+        try:
+            snippets = finding.get_snippets(version_compile.original_sources_path)
+        except:
```
salsolatragus: `get_snippets` throws a `SnippetUnavailableException` in this case, please catch that particular one. This exception knows the file we couldn't extract a snippet from and also the specific error that occurred. Please mention the finding rank and the file in the warning message, to make this more actionable for users. Also please log the full exception for debugging.

mubench.pipeline/tasks/implementations/publish_findings_task.py
```diff
@@ -97,9 +97,15 @@ def __slice_by_max_files_per_post(self, potential_hits: List[SpecializedFinding]
...
-    def _prepare_post(self, finding: SpecializedFinding, version_compile) -> Dict[str, str]:
+    def _prepare_post(self, finding: SpecializedFinding, version_compile, logger) -> Dict[str, str]:
         markdown_dict = self._to_markdown_dict(finding)
-        snippets = finding.get_snippets(version_compile.original_sources_path)
+
+        try:
+            snippets = finding.get_snippets(version_compile.original_sources_path)
+        except SnippetUnavailableException as e:
+            logger.warning("No snippets added for %s", e.file)
+            logger.debug("Failed to find snippets: %s", e.exception)
```
salsolatragus: Isn't `snippets` now undefined in the exception case, leading to an exception in line 109? Can we add a test for this case?

Issue #90, opened 2017-08-18
Recompilation fails on git index


salsolatragus:
Recompilation of a project (`--force-compile`) fails on my machine with and without Docker, because

    Compilation failed: [Errno 13] Permission denied: '/mubench/mubench.pipeline/../checkouts/closure/884/build/.git/objects/pack/pack-cde643dfd753db1fc607c326bcaa140ce157616e.idx'

This is an access-right problem with the git indexes. To delete them manually, I also need to do `rm -f`. Can we work around this?

salsolatragus:
The same happens for SVN projects

    Compilation failed: [Errno 13] Permission denied: '/mubench/mubench.pipeline/../checkouts/jackrabbit/1601/build/.svn/pristine/00/0052b21f676cfd9518da875b7f2bd737600e10c9.svn-base'

PR #126, opened 2017-09-07
Add another fallback for deletion

M8is:
Implementation for #90.

We already had an attempt to work around that problem by setting writable on error. I'm not sure if this double fallback is the way to go, but I couldn't find a better approach. It doesn't seem like there is more we can do on the python side to solve this.

### DIFF
--- a/mubench.pipeline/utils/io.py
+++ b/mubench.pipeline/utils/io.py
@@ -53,7 +53,7 @@ def remove_tree(root: str) -> None:
             try:
                 remove(path)
             except PermissionError:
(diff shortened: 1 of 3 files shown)


mubench.pipeline/utils/io.py
```diff
@@ -52,7 +53,7 @@ def retry_remove(func, path, _):
             try:
                 remove(path)
             except PermissionError:
-                pass
+                Shell.try_exec("rm -f {}".format(path))
```
salsolatragus: We should also quote the path in the command, in case it contains spaces.

Issue #78, opened 2017-08-10
Getting legacy-interface warning twice


salsolatragus:
When I run with an "outdated" detector, I get the warning twice:

```
Starting benchmark...
    The detector uses a legacy interface.
    The following changes were made since its release:
        0.0.10 => 0.0.10:
        Java-side interface now uses builder.
    The detector uses a legacy interface.
    The following changes were made since its release:
        0.0.10 => 0.0.10:
        Java-side interface now uses builder.
Starting requirementscheck at 16:25:17...
```

salsolatragus:
I just realized this also says these are the changes from 0.0.10 to 0.0.10, which should be 0.0.8 to 0.0.10.

salsolatragus:
When changelog is empty or None, nothing should be output.

PR #127, opened 2017-09-07
Legacy warning fix

M8is:
Fix for #78.

Versions are now printed correctly. I couldn't reproduce the duplicate warning though.

### DIFF
--- a/mubench.pipeline/data/runner_interface.py
+++ b/mubench.pipeline/data/runner_interface.py
@@ -1,12 +1,12 @@
 import logging
 from collections import OrderedDict
 from distutils.version import StrictVersion
-from enum import Enum, IntEnum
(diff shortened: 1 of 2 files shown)


mubench.pipeline/data/runner_interface.py
```diff
@@ -198,4 +197,4 @@ def version():
 
     @staticmethod
     def changelog():
-        return ""  # only changed Java-side interface
+        return "Only changed Java-side interface."
```
salsolatragus: I changed this to an empty string on purpose, because there is nothing in that version that impacts the invocation of the detector, hence, no reason to output a warning. This means we need to exclude interfaces with an empty change log when computing whether to output a warning and when computing the warning itself.

Issue #107, opened 2017-09-04
Check validity of violation types in dataset check


salsolatragus:
The dataset check should include whether the entries of the `characterstics` key (list) in `misuse.yml` are valid values. Valid keys, as of now, are:

* `missing/` or `superfluous/`
  * `call`
  * `condition/`
    * `null_check`
    * `synchronization`
    * `value_or_state`
    * `context`
  * `iteration`
  * `exception_handling`
* `misplaced/`
  * `call`

The valid keys can be hardcoded in the respective check.

PR #129, opened 2017-09-07
Check validity of violation types

M8is:
Implementation for #107.


mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -164,6 +165,44 @@ def _location_exists(self, source_base_path, file_, method) -> bool:
         else:
             return len(snippets) > 0
 
+    def _check_violation_types(self, project: Project, misuse: Misuse):
+        file_path = "{}/misuses/{}/misuse.yml".format(project.id, misuse.misuse_id)
+        violation_types = misuse._yaml.get("characteristics", [])
+
+        valid_prefixes = ['missing', 'superfluous', 'misplaced']
+        for violation_type in violation_types:
+            parts = violation_type.split('/')
```
salsolatragus: I would prefer to simplify this by defining an explicit list of all valid values and perform a containment check. This makes it easier to grasp what the valid values are and to change/extend these values in the future.

Issue #146, opened 2017-09-21
Switch from task-wise to project-wise execution


salsolatragus:
It would be much more practical to have the pipeline work project-wise (or even version-wise?), because that would give us intermediate results (e.g., on the review page) before the entire process finishes. Since we don't have the requirements check anymore at the start of each task, there's no reason why we shouldn't change from task-wise to project-wise.

It might then be more practical to chain the tasks as an actual pipeline:

    LoadProjectTask:Project[]
     -> LoadVersionsTask:ProjectVersionData[]
     -> CheckoutTask:ProjectVersionCheckout[]
     -> CompileTask:ProjectVersionCompile[]
     -> DetectTask:DetectorRun[]
     -> Publish:bool

We could the raise exceptions in case of errors, in which case the pipe backtracks to the previous task an continues from there. This might be much simpler than extending the blacklist to exclude projects/versions from subsequent tasks.

salsolatragus:
I pushed the code. The parameter selection based on types is working now. I did not test the error cases yet, to ensure nice error messages if

1. a task requires a parameter that is unavailable,
2. a task returns a result with the same type as an earlier task, and
3. a task declares more than one parameter with the same type (will always have the same value).

Also, we need to handle tasks that raise exceptions when run. Here we should output a respective error message and backtrack, to continue with the next parameter combination.

M8is:
I've implemented the error messages and tests so far.

> Also, we need to handle tasks that raise exceptions when run. Here we should output a respective error message and backtrack, to continue with the next parameter combination.
If a task fails on a certain input, could it return an empty list? This way we would end the recursive chain on that branch and automatically continue with the next set of inputs.

salsolatragus:
We could do this, but why not let the tasks raise an error in case an error occurs and have the handling and reporting implemented once, in the runner? The tasks can still do a silent abort by returning an empty list. The upload tasks would do this, for example, because they have no output. I would like to distinguish failure from no output.

salsolatragus:
Done.

PR #151, opened 2017-10-16
Refactor pipeline

M8is:
Implements #146 

There are some issues in logging/output I would rather discuss with you in person, but everything is running well so far. :)


M8is:
I just noticed some issues with the current white-listing (`--only`). We should probably fix these before merging. Working on it.

mubench.pipeline/tasks/implementations/requirements_check.py
```diff
@@ -2,27 +2,24 @@
 from typing import List
 
 from requirements import Requirement
-from tasks.project_task import ProjectTask
-from utils.shell import Shell
 
 
-class RequirementsCheck(ProjectTask):
+class RequirementsCheck:
```
salsolatragus: I'm not sure whether this should be a task at all... it's not doing anything in run, which indicates it isn't. We could turn in into a proper task, if we allow tasks that just return a `continue`/`stop` symbol, i.e., a kind of control task without output. It appears to me this might make sense in other situations as well (e.g., for the info tasks). What do you think?

mubench.pipeline/tests/tasks/implementations/test_collect_misuses.py
```diff
@@ -0,0 +1,34 @@
...
+from tasks.implementations.collect_misuses import CollectMisuses
+from tests.test_utils.data_util import create_version, create_misuse, create_project
+from utils.data_filter import DataFilter
+
+NO_FILTER = MagicMock()  # type: DataFilter
+NO_FILTER.is_filtered = lambda id_: False
+
+
+class TestCollectMisuses:
+    def test_finds_all_versions(self):
```
salsolatragus: Wrong test name. Copy/paste ;P

mubench.pipeline/tasks/implementations/dataset_check.py
```diff
@@ -33,18 +31,18 @@
 class DatasetCheck(ProjectVersionMisuseTask):
```
salsolatragus: This superclass doesn't exist anymore, right?

mubench.pipeline/tests/tasks/implementations/test_requirements_check.py
```diff
@@ -1,23 +1,23 @@
...
-
-        self.test_requirement = Requirement("-test-requirement-")
-        self.test_requirement.check = MagicMock()
-        self.uut._get_requirements = lambda: [self.test_requirement]
 
+class TestRequirementsCheck:
     def test_passes_on_process_project(self):
-        assert_equals([], self.uut.process_project(create_project("-project-")))
+        self.uut = RequirementsCheck()
+        assert_equals([self.uut], self.uut.run())
```
salsolatragus: This is another indicator to me that this is currently not a task. I get why you are returning `self` here from a technical standpoint, but this is not intuitive from the domain.

mubench.pipeline/tasks/implementations/checkout.py
```diff
@@ -1,54 +1,43 @@
...
-            return self.ok()
+                checkout.create()
 
-        except CommandFailedError as e:
-            logger.error("Checkout failed: %s", e)
-            return self.skip(version)
-        except IOError:
-            logger.error("Checkout failed.", exc_info=True)
-            return self.skip(version)
+        return [checkout]
```
salsolatragus: It somehow bothers me that there are so many tasks returning only a single object and they all have to wrap it in a singleton list... we could automatically wrap non-list values in a singleton list in the runner... what do you think?

mubench.pipeline/utils/data_filter.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+
+class DataFilter:
```
salsolatragus: I'm not entirely happy with how the filtering works... although I'm not sure how a better solution would look like... from the pipeline point of view, we should allow filtering tasks, but this makes things much more complex, since filtering tasks, by definition, return the same type that they consume. To get rid of the duplication in the project/version/misuse tasks, we may be able to implement a filtering wrapper task. What do you think?

mubench.pipeline/tests/tasks/implementations/test_requirements_check.py
M8is: Agreed. We could probably run the check before starting the task runner.

mubench.pipeline/tasks/implementations/checkout.py
```diff
@@ -1,54 +1,43 @@
...
-                    checkout.create()
-
-            return self.ok()
+                checkout.create()
 
-        except CommandFailedError as e:
-            logger.error("Checkout failed: %s", e)
-            return self.skip(version)
-        except IOError:
-            logger.error("Checkout failed.", exc_info=True)
```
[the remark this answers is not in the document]
M8is: The new task runner outputs this as a second debug message for all exceptions [here](https://github.com/stg-tud/MUBench/blob/0341df0dcf8932c03b42a57c4ae6fdd8d328257f/mubench.pipeline/tasks/task_runner.py#L31)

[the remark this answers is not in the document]
salsolatragus: Ok, resolves this for me.

mubench.pipeline/tasks/implementations/collect_misuses.py
```diff
@@ -0,0 +1,10 @@
+from data.project_version import ProjectVersion
+from utils.data_filter import DataFilter
+
+
+class CollectMisuses:
```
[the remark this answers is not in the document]
M8is: We probably still need that. I guess I made a mistake here by oversimplifying this. I'll check it.

mubench.pipeline/tasks/implementations/collect_misuses.py
```diff
@@ -0,0 +1,10 @@
...
+    def __init__(self, data_filter: DataFilter):
+        self.data_filter = data_filter
+
+    def run(self, version: ProjectVersion):
+        return [misuse for misuse in version.misuses if not self.data_filter.is_filtered(misuse.id)]
```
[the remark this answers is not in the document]
M8is: We do when publishing metadata. There's probably a reasonable way around this I guess. It would definitely be nice to get some of that logic out of the data classes.

mubench.pipeline/tasks/implementations/collect_versions.py
```diff
@@ -0,0 +1,10 @@
+from data.project import Project
+from utils.data_filter import DataFilter
+
+
+class CollectVersions:
+    def __init__(self, data_filter: DataFilter):
+        self.data_filter = data_filter
+
+    def run(self, project: Project):
+        return [version for version in project.versions if not self.data_filter.is_filtered(version.id)]
```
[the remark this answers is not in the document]
M8is: Same as with `version.misuses`.

mubench.pipeline/tasks/implementations/requirements_check.py
M8is: For `stop`, we could use the empty list as we do already. Though we could add some constant like `TaskRunner.stop`  to make this more explicit.
`continue` would be nice, even though we would currently only need that for the info tasks.

mubench.pipeline/utils/data_filter.py
M8is: The actual implementation is still buggy. I've spent way too much time to try and make filtering work this way, and I feel like this should actually be handled by the respective tasks. This abstract implementation just doesn't cut it. We should probably discuss this in our meeting.

mubench.pipeline/tasks/implementations/requirements_check.py
salsolatragus: Do we need stop anywhere, where it is not an exception case (such as "unsuccessful compile" or "all projects filtered")? I'm wondering whether stop is always an exception everything else is a continue, either with new data or with the same data.

What does a function return that has no explicit `return`? Is there a special type/object for that? Is it `None`? Would be nice to implement a "continue" task simply by not returning anything...

mubench.pipeline/utils/data_filter.py
salsolatragus: Agreed. Currently the problems are not clear to me, so it's hard to suggest a solution. But if it doesn't work yet, we definitely need another solution ;)

mubench.pipeline/tasks/implementations/requirements_check.py
M8is: Upon implementing I noticed that an explicit `stop` is probably not needed, as tasks use exceptions for that. I would probably return `None` if a task has no result, which allows us to omit the return statement in these tasks.

salsolatragus: This then needs special handling in the runner, right?

    if not result:
      # continue with the same arguments
    else:
      for value in result:
        # continue as before

M8is: Functions without an explicit `return` return `None`. I'll change the implementation to handle this case like a continue. Whenever we `stop`, at least, we probably want to log a `UserWarning` anyway.

salsolatragus: Agreed.

mubench.pipeline/tasks/implementations/collect_versions.py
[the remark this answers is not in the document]
salsolatragus: `project.versions` is indeed only used when uploading metadata (to check whether a misuse is referenced from multiple versions). We could probably replace this usage. We only do this to raise an error, because we currently cannot handle misuses referenced from multiple versions at that point.

mubench.pipeline/tasks/implementations/collect_misuses.py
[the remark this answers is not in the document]
salsolatragus: It's actually also used quite extensively by the `CompileTask` (to access patterns and the misuse source files). Unfortunately, there's no way for a task to request to entire list of values as a parameter and I don't think we can support this, as type hints cannot specify the element type of a `List`... we could replace the usages of `version.misuses` by `CollectMisusesTask().run(version)`... though that's ugly and we'd need the filter for it to work, right?... ideas?

PR #158, opened 2017-11-08
Allow silent stop using empty list as results

### LINKED ISSUE #157: Crash when skipping all versions or misuses
When a task returns an empty list, the task runner fills it with a `continue`. We should not do that, since it leads to unexpected states when using collect tasks. For example, when `CollectVersions` filters all versions, we expect a silent stop. Currently, this will continue without a version available, which for the next task raises an uncaught `TaskParameterUnavailableWarning`.

I suggest we implement an empty results list as a silent stop and only continue, if a task returns `None`.

M8is:
Fix #157.

### DIFF
--- a/mubench.pipeline/tasks/task_runner.py
+++ b/mubench.pipeline/tasks/task_runner.py
@@ -36,7 +36,7 @@ class TaskRunner:
             logger.debug("Full exception:", exc_info=True)
             return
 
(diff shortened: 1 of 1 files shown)


M8is:
Continued in pull request Refactor Task Configuration and Pattern Compile #159

Issue #154, opened 2017-11-02
Pass data-entity list as initial value to runner


salsolatragus:
Invoke runner as `runner.run(data_entity_list)`, instead of passing the list to every collect task's constructor.

Issue #153, opened 2017-11-02
Turn per-misuse compile into pipeline configuration


salsolatragus:
Instead of running per misuse "under the hood" in the run/execution(s), we should make this split on pipeline level.

salsolatragus:
Duplicates #160.

Issue #150, opened 2017-10-02
Compile patterns separately


salsolatragus:
Currently, if two misuses in the same version have a pattern file with the same name, compilation fails, because the files are moved out of the build folder for the first misuse and are then missing for the second. Also, if the two patterns are different, only one of the two versions gets compiled. To avoid such collisions, we should compile patterns separately (which we can do now, since we have the capture the full compile classpath and build the original-jar.

PR #159, opened 2017-11-09
Refactor Task Configuration and Pattern Compile

M8is:
Implements #154 #157 #150 #153


mubench.pipeline/tasks/configurations/configurations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @staticmethod
+    def mode() -> str:
+        raise NotImplementedError
+
+    def _tasks(self, config) -> List:
+        raise NotImplementedError
+
+
+def get_task_configuration(config) -> TaskConfiguration:
```
salsolatragus: We don't really need to pass the `TaskConfiguration` to the outer world, right? All that the caller cares about is a list of tasks. How we realize the mapping from modes to tasks is an implementation detail. Therefore, I would call this function `create_tasks` and return `requested_configuration.tasks()` from it. This would also get us rid of the `self.extend()` call in the constructor, a potential cause of race conditions.

Issue #208, opened 2017-12-14
Hardcode some synthetic hits in DemoDetector


salsolatragus:
To ensure we have some correct findings for E1/3 with the DemoDetector, we should simply hardcode some of the synthetic findings in the detector.

Issue #207, opened 2017-12-13
Finding object has no attribute 'files'


salsolatragus:
We're making progress, but here's the next one:

    Running PublishFindingsTask on <tasks.implementations.findings_filters.PotentialHits object at 0x1110601d0>
            Prepare findings of MuDetectOPAL_171213_woNPE in ex2 for upload to http://mubench2.svamann.de/thesis-rossana/index.php/...
            Preparing findings in project 'closure' version 319...
            Found 9828 potential hits.
            Publishing potential hits...
        'Finding' object has no attribute 'files'

For once it looks like its publishing everything (9828 findings), not just the top ones (this is E2). Should be 50 by default. And then there's the missing `files` property error that prevents the publishing.

PR #211, opened 2017-12-15
Fix publish task

M8is:
Fix _Finding object has no attribute 'files'_ #207
Implement _Hardcode some synthetic hits in DemoDetector_ #208

Specialize findings got lost when we created the filtering tasks. This should fix the problem.

### DIFF
--- a/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/demo/DemoRunner.java
+++ b/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/demo/DemoRunner.java
@@ -35,9 +35,12 @@ public class DemoRunner {
         }
 
         private List<DemoViolation> detectViolations(Path targetSourcePath) throws IOException {
-            return Files.find(targetSourcePath, Integer.MAX_VALUE, this::isJavaFile)
+            List<DemoViolation> violations = Files.find(targetSourcePath, Integer.MAX_VALUE, this::isJavaFile)
                     .map(file -> new DemoViolation(file.toAbsolutePath().toString(), "m()"))
                     .collect(Collectors.toList());
+            // Add synthetic.mapnull.mapnull to ensure we have a potential hit for debugging ex1/3
+            violations.add(new DemoViolation("MapNull.java", "misuse(Map)"));
+            return violations;
         }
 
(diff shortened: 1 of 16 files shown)


mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/demo/DemoRunner.java
```diff
@@ -34,9 +35,11 @@ public DetectorOutput detectViolations(DetectorArgs args) throws Exception {
         }
 
         private List<DemoViolation> detectViolations(Path targetSourcePath) throws IOException {
-            return Files.find(targetSourcePath, Integer.MAX_VALUE, this::isJavaFile)
+            List<DemoViolation> violations = Files.find(targetSourcePath, Integer.MAX_VALUE, this::isJavaFile)
                     .map(file -> new DemoViolation(file.toAbsolutePath().toString(), "m()"))
                     .collect(Collectors.toList());
+            violations.add(new DemoViolation("MapNull.java", "misuse(Map)"));
```
salsolatragus: Please add a comment that states why we add this.

mubench.pipeline/tasks/implementations/findings_filters.py
```diff
@@ -2,13 +2,13 @@
...
-from data.finding import SpecializedFinding, Finding
+from data.finding import Finding
 from data.misuse import Misuse
 
 
 class PotentialHits:
-    def __init__(self, findings: List[SpecializedFinding]):
-        self.findings = findings
+    def __init__(self, findings: List[Finding], detector_run: DetectorRun):
+        self.findings = detector_run.detector.specialize_findings(detector_run.findings_path, findings)
```
salsolatragus: I'm torn whether it makes sense to do the conversion here or in the publish task... after all, we do the conversion for the publication. The currently solution works as well, but it feels a little less explicit. What do you think?

M8is: I've just added it here since it was a small change and the publish findings task seemed pretty loaded already. However, since we've removed it by accident this time... that could probably happen again. I guess it's less likely to be overlooked if we put it where the upload is prepared anyway.

PR #217, opened 2018-01-05
introduce modal for comparing review comments

joschli:
renamed modal for additional run stats and created a new modal for comparing reviews.


salsolatragus:
Formatting of Reviews in model: mockup

mubench.reviewsite/templates/includes/review_modal.phtml
```diff
@@ -0,0 +1,76 @@
...
+                    <?php endforeach; ?>
+                </td>
+            </tr>
+        </table>
+        <h2>Reviews:</h2>
+        <table class="invisible">
+            <tr>
+            <?php foreach($misuse->reviews as $review) : ?>
+            <td>
+                <div class="card-shadow">
```
salsolatragus: Change this to `div` with `float:left` to make it relayout when the window gets too small.

mubench.reviewsite/templates/includes/review_modal.phtml
```diff
@@ -0,0 +1,76 @@
...
+                                        <i class="fa fa-question-circle" aria-hidden="true" style="color:blue;" title="Needs Clarification"></i>
+                                    <?php endif; ?>
+                                </td>
+                    </tr>
+                    <tr>
+                        <td><b>Comment:</b></td>
+
+                    </tr>
+                    <tr><td colspan="2">
+                            <?= $markdown_parser->text($review->comment); ?>
```
salsolatragus: This somehow does not convert newlines into `<br/>`.

joschli: this is standard markdown behavior, only newlines in paragraphs are converted.

salsolatragus: I can't seem to get a line break at all (be it through `<br>` or new `<p>`). How do I start a new paragraph? Blank line doesn't seem to be working. What's the "standard behavior" for this?

joschli: This is an example how it should give you a line break:

"## par 1
This is a line.  
And this is after the break."

I also just found an option to enable normal line breaks.

salsolatragus: Isn't `##` syntax for headers? Anyways, this is certainly too verbose just for a simple line break in a review comment.

Just to make sure we're on the same page: I don't need linebreaks in the form of `<br>`. I want to have *an* intuitive way to get some text to start in a new line. This may as well be through starting a new paragraph. However, this, as well, does not seem to work as expected (through inserting a blank line).

Can you provide a page summarizing the Markdown variant we're using?

joschli: I couldn't really find a page summarizing it. I tested around a little and apparently when you end a line with 2 spaces and then you start a new line it inserts a linebreak.

salsolatragus: * [ ] create markdown parser once, globally
* [ ] enable line breaks on parser

PR #218, opened 2018-01-06
added a page for label management

joschli:


mubench.reviewsite/templates/manage_tags.phtml
```diff
@@ -0,0 +1,70 @@
...
+    <?php foreach($tags as $tag) : ?>
+            <tr>
+                <td>
+                    <input id="tag-input" name="tags[<?= htmlspecialchars($tag->id); ?>][name]" value="<?= htmlspecialchars($tag->name) ?>">
+                </td>
+                <td>
+                    <input id="tag-input" name="tags[<?= htmlspecialchars($tag->id); ?>][color]" value="<?= htmlspecialchars($tag->color) ?>">
+                </td>
+                <td>
+                    <form action="<?= $pathFor('tags.delete', array('tag_id' => $tag->id)) ?>" method="post" target="">
```
salsolatragus: I'm not sure why, but I cannot delete the first tag (Chrome and Safari). The form and button seem to be the same than for the second tag, but somehow the form does not appear in Chrome's DOM hierarchy. Maybe there's a problem with nesting forms that somehow messes up the first nested form? ...

mubench.reviewsite/tests/TagControllerTest.php
```diff
@@ -64,4 +65,18 @@ function test_add_unknown_tag()
...
+        $misuseTags = Misuse::find(1)->misuse_tags;
+        $tag = Tag::where('name', 'test-tag')->first();
+        $this->tagController->updateTag($tag->id, ['name' => 'new-name', 'color' => '#555555']);
+
+        $updatedTag = Tag::where('name', 'test-tag')->first();
+
+        self::assertEquals('new-name', $updatedTag->name);
+        self::assertEquals('#555555', $updatedTag->color);
+    }
 }
```
salsolatragus: Why no test for delete tag? This should ensure that the tag is also removed from any misuse it is linked to.

mubench.reviewsite/setup/create_database_tables.php
```diff
@@ -152,6 +152,7 @@
 Schema::create($tag->getTable(), function (Blueprint $table) {
     $table->increments('id');
     $table->text('name');
+    $table->text('color');
```
salsolatragus: Set length limit.

salsolatragus:
Did you cherry pick the changed from `master-dev` instead of rebasing? There's tons of commits copied in your branch. Please resolve this.

Also:

* Please use the tag colors also on the tag-statistics page (in the headers only).
* Can we do some trick to change the tag-font color back to black for light tag colors? I would guess there's a library function for this somewhere...

joschli:
I just rebased as normal, not sure what caused this.

salsolatragus:
There some of your commits, then those from `master-dev` and then again some of yours. Whatever happen, it certainly wasn't the intended rebase. Anyways, you should be able to resolve this by simply rebasing again on `master-dev`. Since all the intermediate commits *are* on `master-dev`, their duplicates on your branch should become empty and removed in the process. 

I'm a little worried about the conflicts that GitHub shows, since some of these are in files that you did not touch on this branch... if rebasing fails, maybe its easiest to create a new branch from `master-dev` and cherry pick your commits from this branch...

joschli:
Rebasing again solved the strange copied commits. I added a js script to change the color based on the background-color of the tag.

Issue #191, opened 2017-11-30
Convert check tasks to subtasks 


salsolatragus:
Currently, we have `check` and `dataset-check`, I'd like to have something like `check setup` and `check dataset`.

PR #229, opened 2018-01-18
Check subtasks

M8is:
Implements Convert check tasks to subtasks #191


M8is:
I didn't add a description for the `check` subprocess, since I'm not sure how to describe its purpose in a concise way.

PR #231, opened 2018-01-18
Resolve #63: Updating the results for an detector throws error

### LINKED ISSUE #63: Updating results for detector not supported
We do not support updating the results of a detector. Currently, when one does it anyways, the behaviour is strange, because updates go partially through. We should instead give an explicit error message saying that updating is not supported.

joschli:


mubench.reviewsite/tests/RunsControllerTest.php
```diff
@@ -166,18 +166,20 @@ function test_add_run_in_multiple_requests()
...
             "runtime" => 42.1,
             "number_of_findings" => 23,
             "custom2" => "-stat-val2-",
-            "timestamp" => "01-01-2018 01:01:01",
+            "timestamp" => 12,
             "potential_hits" => []
         ];
         $this->runsController->addRun(1, $this->detector1->muid, '-p-', '-v-',  json_decode(json_encode($first_request)));
+        $run = Run::of($this->detector1)->in(Experiment::find(1))->where(['project_muid' => '-p-', 'version_muid' => '-v-'])->first();
+        print_r("BEFORE EXECUTION: " . gettype($run->timestamp));
```
salsolatragus: This is debug output, right? Please remove if so.

mubench.reviewsite/tests/create_simple_test_data.php
```diff
@@ -23,6 +23,7 @@
     $table->float('runtime');
     $table->integer('number_of_findings');
     $table->string('result', 30);
+    $table->integer('timestamp');
```
salsolatragus: I don't like that this script repeats the table creation. Can it use the creation logic from the controllers instead? If you do change this, please ensure it gets into a separate commit!

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -454,6 +455,7 @@ private function createOrUpdateRun(Detector $detector, Experiment $experiment, $
             $savedRun[$column] = $value;
         }
         $savedRun->save();
+        $savedRun = Run::of($detector)->in($experiment)->where(['project_muid' => $projectId, 'version_muid' => $versionId])->first();
```
salsolatragus: Why do we need to fetch the run again?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -238,7 +238,7 @@ function addRun($experimentId, $detectorId, $projectId, $versionId, $run)
 
         if($this->hasRunTable($detector)){
             $saved_run = Run::of($detector)->in($experiment)->where(['project_muid' => $projectId, 'version_muid' => $versionId])->first();
-            if($saved_run->timestamp !== $run->{'timestamp'}){
+            if(intval($saved_run->timestamp) !== $run->{'timestamp'}){
```
salsolatragus: The change below says this is an integer column, so why do you need to cast its value to integer?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -228,23 +233,36 @@ public static function exportRunStatistics($runs)
 
     function addRun($experimentId, $detectorId, $projectId, $versionId, $run)
     {
-        $detector = $this->createDetector($detectorId);
+        $detector = $this->getOrCreateDetector($detectorId);
         $experiment = Experiment::find($experimentId);
 
+        if($this->hasRunTable($detector)){
+            $saved_run = Run::of($detector)->in($experiment)->where(['project_muid' => $projectId, 'version_muid' => $versionId])->first();
+            if(intval($saved_run->timestamp) !== $run->{'timestamp'}){
```
salsolatragus: The table creation says this is an integer column, why do we need to cast it to integer?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -90,9 +90,14 @@ public function postRun(Request $request, Response $response, array $args)
...
             return error_response($response,400, "empty: " . print_r($_POST, true));
         }
+
         $hits = $run->{'potential_hits'};
         $this->logger->info("received data for '" . $experimentId . "', '" . $detector_muid . "." . $project_muid . "." . $version_muid . "' with " . count($hits) . " potential hits.");
-        $this->addRun($experimentId, $detector_muid, $project_muid, $version_muid, $run);
+        try{
+            $this->addRun($experimentId, $detector_muid, $project_muid, $version_muid, $run);
+        }catch(\Exception $e){
+            return error_response($response, 400, $e->getMessage());
```
salsolatragus: I don't really like this handling via exceptions. This would also capture an SQLException, which would not be due to a bad request, but an internal server error... it drops the stacktrace, which could be important in this case. Either we use a dedicated exception type for bad requests and catch it here, or we find some alternative means for error handling without exceptions.  After all, it's not an exceptional case we're treating here, but part of the application logic.

mubench.reviewsite/tests/RunsControllerTest.php
```diff
@@ -135,6 +136,56 @@ function test_store_ex3()
...
+        ];
+        $second_request = [
+            "result" => "success",
+            "runtime" => 42.1,
+            "number_of_findings" => 23,
+            "-custom-stat-" => "-stat-val-",
+            "timestamp" => 14,
+            "potential_hits" => []
+        ];
+        $this->runsController->addRun(1, '-d-', '-p-', '-v-',  json_decode(json_encode($first_request)));
```
salsolatragus: You wanted to check whether we need this encode/decode thing. Do we? Why?

mubench.reviewsite/tests/RunsControllerTest.php
```diff
@@ -135,6 +136,56 @@ function test_store_ex3()
...
+            "result" => "success",
+            "runtime" => 42.1,
+            "number_of_findings" => 23,
+            "custom2" => "-stat-val2-",
+            "timestamp" => 12,
+            "potential_hits" => []
+        ];
+        $this->runsController->addRun(1, $this->detector1->muid, '-p-', '-v-',  json_decode(json_encode($first_request)));
+        $run = Run::of($this->detector1)->in(Experiment::find(1))->where(['project_muid' => '-p-', 'version_muid' => '-v-'])->first();
+        print_r("BEFORE EXECUTION: " . gettype($run->timestamp));
```
salsolatragus: This is debugging output, right? If so, please remove.

mubench.reviewsite/src/Controllers/RunsController.php
joschli: Apparently when using the normal mysql driver in php, it converts all columsn to strings when you select something from the database. This seems to also hold true with our in memory database for tests. Therefore the cast is needed.

Issue #232, opened 2018-01-22
Exclude non-compilable project version per default


salsolatragus:
Currently, when I run the pipeline without specifying a dataset or filter, the pipeline tries to execute on all project versions, even those that do not compile. Since we, by design, fail for non-compilable project versions, it would be nice to avoid the respective error/warning messages. I suggest we simply exclude project versions without `build.commands` in the `version.yml` file.

To include these projects, introduce a `--development-mode`/`-dev` flag.

PR #254, opened 2018-01-29
Implement #232: Exclude non compilable versions per default

M8is:
Use `--development-mode` (abbreviations are enabled, so `-dev` works as well) to disable this feature.

### DIFF
--- a/mubench.pipeline/tasks/implementations/collect_versions.py
+++ b/mubench.pipeline/tasks/implementations/collect_versions.py
@@ -1,17 +1,27 @@
 from data.project import Project
+from data.project_version import ProjectVersion
 from utils.data_entity_lists import DataEntityLists
 
 
 class CollectVersionsTask:
+    def __init__(self, development_mode: bool):
+        self._filter_non_compilable_versions = not development_mode
+
     def run(self, project: Project, data_entity_lists: DataEntityLists):
         return [version for version in project.versions
-                if not self.__is_filtered(project.id, version.id, data_entity_lists)]
+                if not self.__is_filtered(project.id, version, data_entity_lists, self._filter_non_compilable_versions)]
 
     @staticmethod
-    def __is_filtered(project_id: str, version_id: str, data_entity_lists: DataEntityLists) -> bool:
+    def __is_filtered(project_id: str, version: ProjectVersion, data_entity_lists: DataEntityLists,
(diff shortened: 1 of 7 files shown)


mubench.pipeline/tasks/implementations/collect_versions.py
```diff
@@ -1,17 +1,27 @@
...
         return [version for version in project.versions
-                if not self.__is_filtered(project.id, version.id, data_entity_lists)]
+                if not self.__is_filtered(project.id, version, data_entity_lists, self._filter_non_compilable_versions)]
 
     @staticmethod
-    def __is_filtered(project_id: str, version_id: str, data_entity_lists: DataEntityLists) -> bool:
+    def __is_filtered(project_id: str, version: ProjectVersion, data_entity_lists: DataEntityLists,
+                      filter_non_compilable_versions: bool) -> bool:
+        if filter_non_compilable_versions:
+            return not version.is_compilable
```
salsolatragus: Doesn't this disable any black/whitelisting in production mode?

mubench.pipeline/tasks/implementations/collect_versions.py
```diff
@@ -1,17 +1,27 @@
...
 
 
 class CollectVersionsTask:
+    def __init__(self, development_mode: bool):
+        self._filter_non_compilable_versions = not development_mode
+
     def run(self, project: Project, data_entity_lists: DataEntityLists):
         return [version for version in project.versions
-                if not self.__is_filtered(project.id, version.id, data_entity_lists)]
+                if not self.__is_filtered(project.id, version, data_entity_lists, self._filter_non_compilable_versions)]
```
salsolatragus: Why don't we make the private `__is_filtered()` an instance method and access the field directly, instead of passing it around?

mubench.pipeline/tests/tasks/implementations/test_collect_versions.py
```diff
@@ -10,7 +12,7 @@ def test_finds_all_versions(self):
         project = create_project("-project-")
         v1 = create_version("-v1-", project=project)
         v2 = create_version("-v2-", project=project)
-        uut = CollectVersionsTask()
+        uut = CollectVersionsTask(True)
```
salsolatragus: I would prefer the tests to simulate production mode, instead of development mode.

PR #255, opened 2018-01-29
Resolve #210: adding a navbar to the review page.

### LINKED ISSUE #210: Add navigation bar to review page
Add a navigation bar that stays at the bottom of the review page and offers buttons to navigate to the next/previous misuse with potential hits and open review, to save (save and return, save and next), and to return (without saving).

joschli:


salsolatragus:
* [ ] All redirects after save are broken (only path, not base URL).
* [ ] Disable (instead of remove) `>>` and 'Save and next potential hit' if there is no next potential hit.

* [ ] Save + Dropdown -> Dropdown button
* [ ] Make buttons larger and add some padding to the bar

joschli:
This should be ready to merge as well.

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -113,7 +120,8 @@ public function postReview(Request $request, Response $response, array $args)
         $this->updateOrCreateReview($misuse_id, $reviewer->id, $comment, $hits);
 
         if ($review["origin"] != "") {
-            return $response->withRedirect("{$this->settings['site_base_url']}{$review["origin"]}");
+            $base_url = substr($this->settings['site_base_url'], -1) === '/' ? substr($this->settings['site_base_url'], 0, -1) : $this->settings['site_base_url'];
+            return $response->withRedirect("{$base_url}{$review["origin"]}");
```
salsolatragus: So what's the root cause of this '/' problem? I don't like that we're accumulating places where we do this string magic. I think the site_base_url is expected to have a trailing /. Hence, we should generate the origin in a way that does not have a leading /. This should make the string magic here unnecessary.

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -131,7 +139,55 @@ public function updateOrCreateReview($misuse_id, $reviewer_id, $comment, $findin
             $findingReview = FindingReview::firstOrNew(['review_id' => $review->id, 'rank' => $rank]);
             $findingReview->decision = $findings_review['hit'];
             $findingReview->save();
-            $findingReview->violation_types()->sync($findings_review['types']);
+            if($findings_review['types']){
```
salsolatragus: Why is this necessary and what does it have to do with the navigation feature? Doesn't this break the case where I remove all types from a misuse?

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -131,7 +139,55 @@ public function updateOrCreateReview($misuse_id, $reviewer_id, $comment, $findin
...
+    private function collectAllMisuses($runs)
+    {
+        $all_misuses = new Collection;
+        foreach ($runs as $run) {
+            $all_misuses = $all_misuses->merge($run->misuses);
+        }
+        return $all_misuses;
+    }
+
+    private function determineNavigationTargets(Collection $all_misuses, $project_muid, $version_muid, $misuse_muid, $reviewer)
```
salsolatragus: I took the liberty to pull this into a method. Having all this additional logic in getReview certainly made that method to long. Please remember to do such cleanup.

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -131,7 +139,55 @@ public function updateOrCreateReview($misuse_id, $reviewer_id, $comment, $findin
...
+        foreach ($all_misuses as $current_misuse) { /** @var Misuse $current_misuse */
+            if (!$misuse && $current_misuse->getProject() === $project_muid
+                && $current_misuse->getVersion() === $version_muid
+                && $current_misuse->misuse_muid === $misuse_muid) {
+                $misuse = $current_misuse;
+            } else {
+                if ($misuse && !$next_misuse) {
+                    $next_misuse = $current_misuse;
+                }
+                if ((!$next_reviewable_misuse || $misuse)
```
salsolatragus: I'm not sure why this condition is how it is. Please give this a speaking name, either through an intermediate variable or a helper method. Btw. why are there no tests for this logic?

mubench.reviewsite/templates/review.phtml
```diff
@@ -358,10 +360,29 @@ $columns_to_ignore = array(
                     <?php endif; ?>
                     </td></tr>
                 </table>
-                <input type="hidden" name="origin" value="<?= htmlspecialchars($origin_path); ?>">
                 <input type="hidden" name="misuse_id" value="<?= htmlspecialchars($misuse->id); ?>">
-                <input type="submit" value="Save Review" />
-                </form>
+                <div style="position:fixed;bottom:0;left:0;width:100%;background-color:#f1f1f1;z-index:1;padding:5px;">
```
salsolatragus: Please move this styling to the stylesheets.

mubench.reviewsite/templates/review.phtml
```diff
@@ -358,10 +360,29 @@ $columns_to_ignore = array(
...
+                        $next_reviewable_misuse_path = $next_reviewable_misuse ? $pathFor('view', array('experiment_id' => $experiment->id, 'detector_muid' => $detector->muid, 'project_muid' => $next_reviewable_misuse->getProject(), 'version_muid' => $next_reviewable_misuse->getVersion(), 'misuse_muid' => $next_reviewable_misuse->misuse_muid)) . "?origin=$origin_path" . ($experiment->id === 2 ? "&ex2_review_size=$ex2_review_size" : "") : "";
+                        $previous_misuse_path = $previous_misuse ? $pathFor('view', array('experiment_id' => $experiment->id, 'detector_muid' => $detector->muid, 'project_muid' => $previous_misuse->getProject(), 'version_muid' => $previous_misuse->getVersion(), 'misuse_muid' => $previous_misuse->misuse_muid)) . "?origin=$origin_path" . ($experiment->id === 2 ? "&ex2_review_size=$ex2_review_size" : "") : "";
+                        ?>
+                    <a href="<?= $origin_path ?>" style="padding:5px;text-align: center;" class="button">return</a>
+                    <a href="<?= $previous_misuse_path ?>" style="padding:5px;text-align: center;" class="button <?= $previous_misuse_path ? "" :  "disabled" ?>"><</a>
```
salsolatragus: I understand that the placement of the buttons is specific to this navigation bar, but why do you add the padding here as well? Shouldn't this be part of the class? (the question applies to all elements in this navigation bar)

salsolatragus:
The `return` button breaks when there is a URL prefix.

salsolatragus:
Sometimes the origin gets messed up and than we either miss the URL prefix or have it twice:

* When I open a review, I have `origin=/private/experiments/3/detectors/DMMC/runs` and the `return` button is broken. `Save` and `Save and return` work, though.
* When I open a review and save it, there's no origin and `return` works. `Save and return` is broken then, though. `Save` still works.

mubench.reviewsite/src/Controllers/ReviewsController.php
joschli: next_reviewable_misuse is null before going through all the misuses. We set it on the first available misuse that is reviewable or on the first reviewable misuse after we already found the current misuse.

joschli:
What url does the return button link in the first case to ? It should link to `/private/experiments/3/detectors/DMMC/runs`.
Funny enough both cases work on my test instance... Can you check what path is saved for origin in the js script in the case you have no origin set in the url ?

mubench.reviewsite/src/Controllers/ReviewsController.php
salsolatragus: Oh, I get it. So this means `if (isFirstReviewableMisuse($current_misuse) || isFirstReviewableMisuseAfter($current_misuse, $misuse))`. Make sense, but did not reveal itself to me. Please use some intermediate variables or helper functions to give names to these intentions.

salsolatragus:
You can play around with this yourself: http://mubench.svamann.de/fse18-ext/private/experiments/3/detectors/DMMC/runs

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+                        "rank" => 0,
+                    ],
+                    [
+                        "misuse" => "2",
+                        "rank" => 0,
+                    ]
+                ]
+            ]
+        )));
+        Misuse::create(['misuse_muid' => '1', 'run_id' => 1, 'detector_id' => 1]);
```
salsolatragus: This fails, because the misuse with `muid` 1 has been created by `addRun` already. Same goes for the creation in the next line.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+        list($previous_misuse, $next_misuse, $next_reviewable_misuse, $misuse)  = $this->reviewController->determineNavigationTargets(Misuse::all(), '-p-', '-v-', '0', Reviewer::find(1));
+        self::assertEquals('2', $previous_misuse->misuse_muid);
+        self::assertEquals('1', $next_misuse->misuse_muid);
+        self::assertEquals('2', $next_reviewable_misuse->misuse_muid);
+        self::assertEquals('0', $misuse->misuse_muid);
+    }
+
+    function test_determine_next_is_previous_misuse()
+    {
+        $this->setupSomeRunWithThreeMisusesAndTwoReviews();
```
salsolatragus: This helper suggests that it is irrelevant which misuses you create and which of them have reviews. In fact, the name is even wrong, because you're creating three misuses two of which have two reviews each. Regardless, it is very important for the test which misuses have how many reviews, because that determines the outcome of the logic (w.r.t next reviewable). Therefore, please explicitly create the misuses and the reviews in the test methods. I'm sure you don't actually need the same data setup in both tests.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+
+        list($previous_misuse, $next_misuse, $next_reviewable_misuse, $misuse)  = $this->reviewController->determineNavigationTargets(Misuse::all(), '-p-', '-v-', '2', Reviewer::find(1));
+        self::assertEquals('1', $previous_misuse->misuse_muid);
+        self::assertEquals('1', $next_misuse->misuse_muid);
```
salsolatragus: Next would be 0 here, not 1.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+        self::assertNull($next_reviewable_misuse);
```
salsolatragus: I'm not sure why there would be no reviewable, because misuse with `muid` 0 has no reviews... and current is 2.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
         self::assertEquals("missing/call", $violation_types[0]["name"]);
     }
 
+    function test_determine_next_misuse()
```
salsolatragus: This is a bad name. The test is asserting much more than the next element.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+    {
+        $this->setupSomeRunWithThreeMisusesAndTwoReviews();
+
+        list($previous_misuse, $next_misuse, $next_reviewable_misuse, $misuse)  = $this->reviewController->determineNavigationTargets(Misuse::all(), '-p-', '-v-', '0', Reviewer::find(1));
+        self::assertEquals('2', $previous_misuse->misuse_muid);
```
salsolatragus: Please save the misuses in variables and use their muids as the expected values here, otherwise one has to trace the ids to check what this means.

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+                    [
+                        "misuse" => "2",
+                        "rank" => 0,
+                    ]
+                ]
+            ]
+        )));
+        Misuse::create(['misuse_muid' => '1', 'run_id' => 1, 'detector_id' => 1]);
+        Misuse::create(['misuse_muid' => '2', 'run_id' => 1, 'detector_id' => 1]);
+        Reviewer::create(['name' => 'reviewer2']);
```
salsolatragus: Please save the reviewer(s) in variables and pass their ids on, otherwise it is hard to track which reviewer has which id and you're relying on the database generating keys in insertion order!

mubench.reviewsite/tests/ReviewControllerTest.php
```diff
@@ -54,4 +54,60 @@ function test_stores_violation_types()
...
+        $this->reviewController->updateOrCreateReview(1, 1, '-comment-', [['hit' => 'No', 'types' => [1]]]);
```
salsolatragus: You could create a helper method `createConclusiveReviewState($misuse, $reviewer1, $reviewer2)`, because the details of how this state is constructed are indeed irrelevant for the tests.

salsolatragus:
Our fix from yesterday broke the return path after saving a snippet. It now has a duplicated url prefix.

joschli:
I fixed the problem with saving snippets. 
I revised the ReviewController tests to make the test situation clear.
When just saving a review the origin is now preserved.

salsolatragus:
"Save and next" sends me literally to the next misuse (whether this one has potential hits or not), as opposed to the next misuse with a missing review. The button is enabled/disabled correctly according to whether there is another reviewable misuse. Please change the destination of "Save and next" for this to be consistent.

salsolatragus:
Pass origin through snippet add/delete.

joschli:
This should be ready to merge. 
The Redirect when deleting/adding snippets/tags now preserves the origin.

PR #259, opened 2018-01-30
Resolve #215: redirecting to setup when no experiment table exists.

### LINKED ISSUE #215: Auto-forward to setup
Can we make the review site forward to the setup automatically, if no database tables exist? Ideally, we could check whether there is **any** table with the configured table prefix, because that would be independent of the concrete table names. However, we could also pick a table name where we are reasonably sure it will not change.

The setup should automatically redirect back to the page, on success.

joschli:
I also added a redirect to the index of the reviewsite when the setup script is completed.


salsolatragus:
Also, we should check whether the `setup.php` script exists and `die("No database setup and setup script not found.")` otherwise, to avoid redirecting into a 404.

joschli:
When does the sql error occur for you ? When you delete your test db completely or when you change to a new prefix ?

joschli:
I should have fixed the issue by moving the check back into the bootstrap script. The error should come from the renderer setup where we already access the database.

Issue #239, opened 2018-01-25
Provide Classpaths as Objects


salsolatragus:
Currently, we split the classpath and provide the entries as an array of strings. I noticed, however, that most of the current detectors actually use it as a classpath, i.e., they join in back together. Therefore, I think it would be a great idea to provide the classpath as a `ClassPath` object that can return either the entire classpath as a string or the individual paths in it.

Also it would be nice to have provide an `append(ClassPath)` and an `append(String entry)` method on `ClassPath`, to allow constructing larger classpaths.

PR #263, opened 2018-01-31
Implement  #239: Provide Classpaths as Objects

M8is:
The CLI version is not updated since I didn't know if you want a release for this. Didn't add a new detector interface on the python side for the same reason.

We're handling empty inputs as well. This might be overkill, but I thought it would be better to make sure nothing weird happens.

As discussed, we only provide the class paths as `ClassPath`. Source paths are still String arrays.

### DIFF
--- a/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/ClassPath.java
+++ b/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/ClassPath.java
@@ -1,7 +1,7 @@
 package de.tu_darmstadt.stg.mubench.cli;
 
 public class ClassPath {
-    private String path;
+    private final String path;
 
     public ClassPath(String path) {
         this.path = path;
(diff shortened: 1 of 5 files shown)


M8is:
Implements Provide Classpaths as Objects #239.

mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/ClassPath.java
```diff
@@ -0,0 +1,33 @@
...
+        } else {
+            return new String[0];
+        }
+    }
+
+    public String getClasspath() {
+        return path;
+    }
+
+    public void append(ClassPath other){
```
salsolatragus: I would prefer `ClassPath` to be a value object (make `path` final) and `append()` to return a new `ClassPath` instance. Otherwise people may modify the inputs given to the detector, which may cause confusion. I see that `append` may not be an ideal name then, because it suggests modification (as for lists). Do you have a suggestion for a better name? (join, plus, ...)

mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/ClassPath.java
```diff
@@ -0,0 +1,33 @@
...
+        if (other.getClasspath() != null && !other.getClasspath().isEmpty()) {
+            this.path += ":" + other.getClasspath();
+        }
+    }
+
+    public void append(String entry) {
+        if (entry != null && !entry.isEmpty()) {
+            this.path += ":" + entry;
```
salsolatragus: Handle `this.path.isEmpty()`. Above as well.

PR #279, opened 2018-02-05
Skip files on conflict when creating jar archive

### LINKED ISSUE #257: Compiling Apache Gora prints warning
When I compile `apache-gora`, I get multiple warnings from the creation of the `original-classes.jar`, because multiple files with the same name get added to the archive.

    Compiling project 'apache-gora' version bb09d89...
    /usr/lib/python3.6/zipfile.py:1349: UserWarning: Duplicate name: 'gora.properties'
      return self._open_to_write(zinfo, force_zip64=force_zip64)
    /usr/lib/python3.6/zipfile.py:1349: UserWarning: Duplicate name: 'META-INF/LICENSE'
      return self._open_to_write(zinfo, force_zip64=force_zip64)
    /usr/lib/python3.6/zipfile.py:1349: UserWarning: Duplicate name: 'META-INF/DEPENDENCIES'
      return self._open_to_write(zinfo, force_zip64=force_zip64)
    /usr/lib/python3.6/zipfile.py:1349: UserWarning: Duplicate name: 'META-INF/NOTICE'
      return self._open_to_write(zinfo, force_zip64=force_zip64)

While this does not seem to be a problem (the Jar is created successfully), its at least confusing to the user. Can we handle this? I.e., suppress it for any non .class files and raise an actual error should this ever happen for a .class file (which might mean we create a broken Jar)?

M8is:
Resolve #257: Compiling Apache Gora prints warning

### DIFF
--- a/mubench.pipeline/utils/io.py
+++ b/mubench.pipeline/utils/io.py
@@ -1,3 +1,4 @@
+import logging
 import zipfile
 from os import makedirs, chmod, remove, listdir, readlink, symlink, stat, walk
 from os.path import dirname, exists, isfile, join, isdir, basename, islink, relpath
@@ -157,4 +158,8 @@ def zip_dir_contents(sources: List[str], destination: str):
                 for file in files:
                     file_path = join(root, file)
                     file_path_in_archive = relpath(file_path, source)
-                    archive.write(file_path, file_path_in_archive)
+                    if not any(file.endswith(file_path_in_archive) for file in archive.namelist()):
+                        archive.write(file_path, file_path_in_archive)
+                    else:
+                        logger = logging.getLogger("io.zip_dir_contents")
+                        logger.debug("File conflict in zip {}: {}".format(destination, file_path_in_archive))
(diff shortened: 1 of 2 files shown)


mubench.pipeline/utils/io.py
```diff
@@ -157,4 +158,8 @@ def zip_dir_contents(sources: List[str], destination: str):
                 for file in files:
                     file_path = join(root, file)
                     file_path_in_archive = relpath(file_path, source)
-                    archive.write(file_path, file_path_in_archive)
+                    if not any(file.endswith(file_path_in_archive) for file in archive.namelist()):
```
salsolatragus: Why is this `endswith`? Isn't a conflict only when they are equal? This way, if the archive contains `foo/Bar.java` adding `Bar.java` would be rejected, wouldn't it?

salsolatragus: Why is this `endswith`? Isn't a conflict only when they are equal? This way, if the archive contains `foo/Bar.java` adding `Bar.java` would be rejected, wouldn't it?

PR #283, opened 2018-02-06
Make detector tags case insensitive

### LINKED ISSUE #250: Make detector tags case insensitive
Currently, detector tags are case sensitive. I would prefer if they were insensitive. We can do this by simply converting them always to lower case before use. This includes both the tag passed to the command line, as well as the tags read from the `releases.yml` files.

M8is:
Resolve #250: Make detector tags case insensitive.

### DIFF
--- a/mubench.pipeline/tests/data/test_detector.py
+++ b/mubench.pipeline/tests/data/test_detector.py
@@ -79,7 +79,7 @@ class TestDetector:
             {"md5": "-md5_3-", "tag": "-release_3-", "cli_version": "-version-"}])
         detector = Detector(self.temp_dir, self.detector_id, [], "Release_Requested")
 
-        expected_url = "{}/RELEASE_REQUESTED/{}/{}.jar".format(Detector.BASE_URL,
+        expected_url = "{}/release_requested/{}/{}.jar".format(Detector.BASE_URL,
                                                                RunnerInterfaceTestImpl.TEST_VERSION, self.detector_id)
         assert_equals(expected_url, detector.jar_url)
         assert_equals("-md5_requested-", detector.md5)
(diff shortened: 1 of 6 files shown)


mubench.pipeline/tests/data/test_detector.py
```diff
@@ -71,6 +71,19 @@ def test_raises_on_no_matching_release(self):
...
 
+    def test_release_is_case_insensitive(self):
+        self.setup_releases([
+            {"md5": "-md5_1-", "tag": "-release_1-", "cli_version": "0.0.0"},
+            {"md5": "-md5_requested-", "tag": "-Release_requested-",
+             "cli_version": RunnerInterfaceTestImpl.TEST_VERSION},
+            {"md5": "-md5_3-", "tag": "-release_3-", "cli_version": "0.0.2"}])
+        detector = Detector(self.temp_dir, self.detector_id, [], "-release_Requested-")
+
+        expected_url = "{}/-Release_requested-/{}/{}.jar".format(Detector.BASE_URL,
```
salsolatragus: The URL should use the lower case variant, too.

mubench.pipeline/tests/data/test_detector.py
```diff
@@ -71,6 +71,19 @@ def test_raises_on_no_matching_release(self):
         self.setup_releases([{"md5": "-md5-", "tag": "-release-", "cli_version": "0.0.1"}])
         assert_raises(ValueError, Detector, self.temp_dir, self.detector_id, [], "-unavailable_release-")
 
+    def test_release_is_case_insensitive(self):
```
salsolatragus: Please make the differences in casing more obvious, like define 'TAG', request 'TaG', URL 'tag'. Also drop the '-' around the tag, because our convention is that these express "this string is an arbitrary-but-fixed value" where in this case is actually is not.

The 'cli_version' on the other hand, could be an arbitrary string here.

salsolatragus:
Please also add a note to the CLI `--tag` option and convert values to lower case there, like you do for dataset names.

Issue #240, opened 2018-01-25
Provide utility to convert method names to MUBench format


salsolatragus:
We have detectors that obtain method names in [JVM format](https://docs.oracle.com/javase/specs/jvms/se7/html/jvms-4.html#jvms-4.3.3) and in an qualified representation, e.g., `org.apache.jackrabbit.spi.commons.value.QValueFactoryImpl$BinaryQValue.getStream () : java.io.InputStream`. It would be nice to have some utility that can automatically extract the MUBench method signature from these representations. This utility should be able to

* [x] normalize JVM primitive type names, e.g., `I`
* [x] reduce parameter type names to their simple name
* [x] remove return types
* [x] ensure parameters are separated by `, ` (comma, space)
* [x] replace `<init>`/`<cinit>` method names by the declaring classes simple name
* [x] extract the qualified name of the declaring type, if possible.

PR #307, opened 2018-02-27
Provide utility to convert method names to MUBench format

M8is:
Implement #240: Provide utility to convert method names to MUBench format.

Took the MuBenchMethodFormatConverter from the FindBugs project on Bitbucket. Since the JVM representation doesn't contain the additional information, we need to pass things like the declaring type name, and method name manually.

### DIFF
--- a/mubench.cli/pom.xml
+++ b/mubench.cli/pom.xml
@@ -4,7 +4,7 @@
 
 	<groupId>de.tu-darmstadt.stg</groupId>
 	<artifactId>mubench.cli</artifactId>
-	<version>0.0.12-SNAPSHOT</version>
+	<version>0.0.13-SNAPSHOT</version>
 	<packaging>jar</packaging>
 
 	<name>mubench.cli</name>
--- a/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/MuBenchMethodName.java
+++ b/mubench.cli/src/main/java/de/tu_darmstadt/stg/mubench/cli/MuBenchMethodName.java
@@ -6,45 +6,60 @@ import java.util.regex.Pattern;
 
 public class MuBenchMethodName {
     private final String parameterPattern = "\\(.*?\\)";
-    private final String parameterSeparatorPattern = "(/|,)";
+    private final String parameterSeparatorPattern = "([/,])";
 
-    private final String name;
-    private final String declaringTypeName;
+    private final String originalName;
 
     public MuBenchMethodName(String name) {
-        this.name = convertToMuBenchName(name);
-        this.declaringTypeName = getDeclaringTypeName(name);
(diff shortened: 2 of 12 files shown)


salsolatragus:
Check whether the utils work for the following detector runners:

* [ ] DMMC
* [ ] Jadet/Tikanga

salsolatragus:
What's the state of this? Can I support this?

M8is:
Currently, I checked for compatibility with Jadet and Tikanga. It seems to work well for these, but I noticed that my naming of `SourceCodeMethodIdentifier` is probably way off there. @salsolatragus You could suggest another name.
I noticed that DMMC uses yet another format, but it since the conversion is basically already implemented there, it shouldn't be a problem to reuse that code for a `MethodIdentifier` implementation. I can probably do that before thursday.

salsolatragus:
`DMMC` is using plain Bytecode representation, as far as I remember, doesn't it? Didn't you implement that already? `JVMMethodIdentifier` or `BytecodeMethodIdentifier` would be options then.

The usage representation of Jadet and Tikanga is called OUM (Object Usage Model). If the implementation is indeed specific to these two detectors, we could call it `OUMMethodIdentifier`?

I remember that I thought we could have a more general identifier that happens to also properly handle OUM identifiers... might have been wrong though. If the implementations are indeed specific to certain detectors, I would agree to place them in the runners, even though we then still have duplication between Jadet and Tikanga. At least the `JVMMethodIdentifier` should be in `cli` though, since it's independent of a concrete detector.

M8is:
`DMMC` uses `ObjectTrace` instances, which contain a `location` (the class as a FQN) and a context (the method signature). If we combine these, we should be able to reuse the `OUMMethodIdentifier` used for `Jadet` and `Tikanga`. 
I've implemented this in `DMMCMethodIdentifier`, so you can take a look. I'd argue that it would be best if `DMMCMethodIdentifier` could use `ObjectTrace` directly, so I'd probably move it into `DMMC`.

M8is:
About where to put the `MethodIdentifier` implementations before we do that: I think `OUMMethodIdentifier` and `JVMMethodIdentifier` are fine, and with the `DMMCMethodIdentifier` out, we'd be clean.

What's left to do is to release the `mubench.cli` changes so we can officially use these identifiers in the detectors.

salsolatragus:
Yes, we need to release. Can do this on Thursday.

I agree with your placement of the classes, but I'm now again unhappy with the `OUM` name, because `DMMC` does not use OUM... the question is, what is it then?

M8is:
`Jadet`, `Tikanga`, and `DMMC` now use the new util.

salsolatragus:
Ok, this means I can merge this PR and finally work on closing #239, right?

Issue #294, opened 2018-02-09
Get rid of detectors.py


salsolatragus:
I want to get rid of the `detectors.py` files, because their management is a hassle and the reflection loading is a unnecessarily complex part of the pipeline. I would like the functionality to move into the runners. Currently, really all the `detectors.py` does is type conversion, rounding, and conversion of dot to images. The former two can be performed by the runner themselves as well. The dot-conversion should not, because the runner would need to create all possible dots, which is often unnecessary. Therefore, we should provide some `DotGraph` entity with the `mubench.cli` that runners wrap the dot-String in and that indicates to the pipeline that this field needs to be converted to an image before publishing. This means that instead of invoking the specialization on the detector, we would check every entry of the finding for such a flag an apply the respective transformation automatically, if needed.

PR #310, opened 2018-02-28
Get rid of detectors.py

M8is:
Implement #294: Get rid of detectors.py

@salsolatragus This isn't finished yet, but I'd like to have your opinion. 

The specialization is now done by the pipeline, making the `Detector` subclasses obsolete. I've even disabled the look-up, and we always use the base `Detector`.

As for graphs, we use the `digraph ...{...}` structure to automatically convert them. I think it makes sense to use the string's structure in this case as it is very specific. Also, this should be backwards compatible if I'm not mistaken. 

We lose some features though, like float formatting. Could a detector add a float to the finding? That way, we could apply rounding on all floats and let detectors report strings if they require an accurate value for some reason.

Another feature we lose is that DMMC reports a `firstcallline` instead of `startline`, which we patch in the specialization. In my opinion, this should be a problem for the detector to fix.


mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -139,3 +130,20 @@ def __get_publish_findings_url(self, detector, project, version):
         experiment_id = self.experiment_id[2:]  # the review site uses only the experiment number as Id
         return urljoin(self.review_site_url, "experiments/{}/detectors/{}/projects/{}/versions/{}/runs".format(
             experiment_id, detector.id, project.id, version.version_id))
+
+    @staticmethod
+    def _convert_graphs_to_files(potential_hit: Dict, findings_path: str) -> List[str]:
+        files = []
+
+        graph_pattern = re.compile("digraph .* {.*?}", re.DOTALL)
```
salsolatragus: Consider all graph formats: https://en.wikipedia.org/wiki/DOT_(graph_description_language)

M8is:
Done a lot of cleanup. 
Also, we now load detectors in subfolders per, which allows the user to load multiple different tags of the detector. `latest` is still considered default, but will use the top listed release, so nothing has changed in that regard.
@salsolatragus Should be ready for another review.

PR #313, opened 2018-03-01
Refactor tests and controllers to use arrays instead of json objects.

joschli:


mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -415,14 +415,15 @@ function addColumnToTable($table_name, $column)
     private function storeFindings(Detector $detector, Experiment $experiment, $projectId, $versionId, Run $run, $findings)
     {
         foreach ($findings as $finding) {
-            $misuseId = $finding->{'misuse'};
+            $finding = (array) $finding;
```
salsolatragus: Why cast?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -86,12 +86,12 @@ public function postRun(Request $request, Response $response, array $args)
         $detector_muid = $args['detector_muid'];
         $project_muid = $args['project_muid'];
         $version_muid = $args['version_muid'];
-        $run = decodeJsonBody($request);
+        $run = decodeJsonBody($request, true);
```
salsolatragus: Remove `false` option.

Issue #271, opened 2018-02-02
Make misuse locations more precise


salsolatragus:
Currently, we locate misuses by the file and the method signature. Now we have multiple misuses in the dataset where we find multiple classes in the respective file, each of which has a method matching the signature. Hence, MUBench cannot decide which of these methods we're talking about and conservatively assumes it might be any of them. I would like to improve this.

The problem is we need a way to designate the location that can be determined from both source code and Bytecode, in order for all detectors to be able to do it. On the other hand, we can assume that we always refer to a fixed snapshot of the project, i.e., we can use line numbers, since they will never change in our scenario.

Consider the following example, which illustrates all the different cases I can think of:

    01: // org/foo/Bar.java
    02: package org.foo;
    03: 
    04: class Bar {
    05:   X m() { ... } // (A) method in primary class
    06: 
    07:   class Blub {
    08:     X m() { ... } // (B) method in inner class
    09:   }
    10: 
    11:   Y f() {
    12:     new a.b.C() {
    13:       X m() { ... } // (C) method in anonymous class instance
    14:     }
    15:   }
    16: }
    17: 
    18: class Baz {
    19:   X m() { ... } // (D) method in secondary class
    20: }

**Solution using line numbers**

Probably the most general way to do this is to use the file name and the number of the first line of the method declaration:

(A) `org/foo/Bar.java`, 05
(B) `org/foo/Bar.java`, 08
(C) `org/foo/Bar.java`, 13
(D) `org/foo/Bar.java`, 19

However, I'm not sure how easy it is to retrieve this information from Bytecode, especially in the latter three cases, where the classes actually get compiled into their own class files.

**EDIT**: [Source line information for method declarations should be available in the Bytecode](https://stackoverflow.com/a/7485720/1372671).

**Solution using type names**

We could also approach this using the qualified names:

(A) `org/foo/Bar.java`, `org.foo.Bar`, `m()`
(B) `org/foo/Bar.java`, `org.foo.Bar$Blub`, `m()`
(D) `org/foo/Bar.java`, `org.foo.Baz`, `m()`

But this is difficult for anonymous class instances, since their names are generated during compilation, i.e., we cannot use them for localization in the source. Since multiple anonymous class instances of the same type may be instantiated in the same method, the location of the enclosing method is not unique either. Hence, we would need to fall back to line numbers here, I guess:

(C) `org/foo/Bar.java`, 12:`a.b.C`, `m()`

PR #316, opened 2018-03-02
Make misuse locations more precise

M8is:
Implement #271: Make misuse locations more precise

@salsolatragus I'm not sure what key you would like to use for `startline`, but other than that this should work now. The other question is: do you want it to be added to locations or not? I think it makes sense, but it seems to add that to the upload as well. I don't know if that's a problem.

### DIFF
--- a/mubench.pipeline/data/misuse.py
+++ b/mubench.pipeline/data/misuse.py
@@ -11,16 +11,16 @@ from data.snippets import get_snippets, Snippet
 
 
 class Location:
-    def __init__(self, file: str, method: str, startline: int):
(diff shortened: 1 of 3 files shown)


salsolatragus:
I started adding this in the data as a `line` key below `location`, like [here](https://github.com/stg-tud/MUBench/blob/master-dev/data/asterisk-java/misuses/dmmc-8/misuse.yml#L13). I think this makes sense.

It's not a problem that this information is sent to the review site. Currently, it is just dropped there, but ultimately I would like to store and display it there as well (#317).

M8is:
Ready to review.

Issue #238, opened 2018-01-24
Publish only run stats in Experiment 2


salsolatragus:
I want the possibility to publish only the run statistics in Experiment 2. We need this to sample the projects we review, because we look at the number of findings the detectors have. On pipeline side we can probably do this by setting `--limit 0`. The review site needs to accept an update of a run, if the previous one had 0 (uploaded) findings.

salsolatragus:
The pipeline part is implemented. The review-site doesn't need to change, because it will implicitly accept updates after we merged #331.

PR #330, opened 2018-03-19
Implement #238: Publish only run stats in Experiment 2

M8is:
Implement #238: Publish only run stats in Experiment 2 

We can now use `--limit 0`, which should filter all findings. I've also removed default handling in the filter, as we used a default of 50 and the special case for 0 was not documented anyway. In my opinion, it makes sense to always limit the potential upload size, even if the limit is very high.

I've also changed the output a bit to make this more clear. Here's an example:
```
> ./mubench publish ex2 MuDetect --limit 0 -s http://localhost --only bcel
Starting benchmark...
        All requirements satisfied. You're good to go.
        Compiling project 'bcel' version 24014e5...
        Successful previous run on project 'bcel' version 24014e5. Skipping.
        Detector reported 316 findings.
            Publishing findings of MuDetect in ex2 on project 'bcel' version 24014e5 for upload to http://localhost...
            Uploading 0 potential hits.
            ...
```

### DIFF
--- a/mubench.pipeline/data/detector_run.py
+++ b/mubench.pipeline/data/detector_run.py
@@ -80,6 +80,7 @@ class DetectorRun:
             return
         elif self.is_success():
             logger.info("Successful previous %s. Skipping.", self)
+            logger.info("Detector reported %s findings.", len(self.findings))
             return
 
         self.reset()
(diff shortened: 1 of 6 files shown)


M8is:
Isn't the semantics as expected now? `--limit 0` means limit the upload to 0 findings, just as any other number would. But you're right, we should add a hint that this can be used to publish the run stats exclusively.

M8is:
Ready for another review.

Issue #194, opened 2017-12-01
Update run data on republish before any reviews


salsolatragus:
The review site should allow updates of run data as long as there are no reviews for that run. In the simplest case we can detect an update, because the run result changed, e.g., from "error" to "success". However, to detect that only the findings changed, we'd probably have to introduce a run timestamp on pipeline side and send it to the review site.

PR #331, opened 2018-03-19
Implement #194: Update run data on republish before any reviews

M8is:
Implement #194: Update run data on republish before any reviews

@salsolatragus I'm not sure if this is what you meant. We now use the run timestamp for uploads.

### DIFF
--- a/mubench.pipeline/data/detector_run.py
+++ b/mubench.pipeline/data/detector_run.py
@@ -29,38 +29,39 @@ class DetectorRun:
         self.findings_path = findings_path
         self.__FINDINGS = None
         self.__POTENTIAL_HITS = None
-        self.__run_info = None
+        self.__RUN_INFO = None
         self._findings_file_path = join(findings_path, DetectorRun.__FINDINGS_FILE)
         self._run_file_path = join(findings_path, DetectorRun.__RUN_FILE)
 
-    def __get_run_info(self, key: str, default):
-        if not self.__run_info:
-            self.__run_info = self.__load_run_info()
-        return self.__run_info.get(key, default)
+    @property
+    def __run_info(self):
+        if not self.__RUN_INFO:
+            self.__RUN_INFO = self.__load_run_info()
+        return self.__RUN_INFO
 
     def __load_run_info(self):
         return read_yaml_if_exists(self._run_file_path)
 
     @property
     def result(self):
-        result = self.__get_run_info("result", None)
+        result = self.__run_info.get("result", None)
         return Result[result] if result else None
 
     @property
     def runtime(self):
-        return self.__get_run_info("runtime", 0)
+        return self.__run_info.get("runtime", 0)
 
     @property
     def message(self):
-        return self.__get_run_info("message", "")
+        return self.__run_info.get("message", "")
 
     @property
     def __detector_md5(self):
-        return self.__get_run_info("md5", None)
+        return self.__run_info.get("md5", None)
 
     @property
     def __timestamp(self):
-        return self.__get_run_info("timestamp", 0)
+        return self.__run_info.get("timestamp", 0)
 
     def get_run_info(self):
         run_info = {
@@ -137,7 +138,7 @@ class DetectorRun:
             "md5": detector_md5
         })
         write_yaml(run_info, file=self._run_file_path)
-        self.__run_info = run_info
+        self.__RUN_INFO = run_info
 
     @property
     def findings(self) -> List[Finding]:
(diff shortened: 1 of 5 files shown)


mubench.pipeline/tests/data/test_detector_run.py
```diff
@@ -183,3 +183,13 @@ def test_run_outdated_on_newer_compile(self, read_yamls_mock, _):
...
+
+    @patch("data.detector_run.DetectorRun._load_findings")
+    @patch("data.detector_run.open_yamls_if_exists")
+    def test_get_run_info_without_previous_load(self, read_yamls_mock, load_findings_mock, _):
+        read_yamls_mock.return_value = {}
+        load_findings_mock.return_value = []
+
+        uut = DetectorRun(self.detector, self.version, self.findings_path)
+
+        run_info = uut.get_run_info()
```
salsolatragus: I'm not sure what the purpose of this test is. There's no assertion?

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -92,7 +91,7 @@ def __to_postable_data(self, run_info, result, postable_potential_hits: List[Dic
         data = self._to_markdown_dict(run_info)
         data["result"] = result
         data["potential_hits"] = postable_potential_hits
-        data["timestamp"] = self.run_timestamp
+        data["timestamp"] = run_info["timestamp"]
```
salsolatragus: We don't need to explicitly copy the timestamp. The entire `run_info` is copied into the `data` three lines above.

mubench.pipeline/data/detector_run.py
```diff
@@ -66,6 +66,7 @@ def get_run_info(self):
         run_info = {
             "number_of_findings": self.number_of_findings,
             "runtime": self.runtime,
+            "timestamp": self.__timestamp
```
salsolatragus: Do we really need to do this? I thought that we store the run timestamp in the run info already, to trigger detector rerun if the existing run is older than the current compile? Or don't we do this?

mubench.pipeline/tests/data/test_detector_run.py
M8is: There's a `dict.update(None)` happening, and I wasn't sure if that even works. In actual runs, the internal `self.__run_info` would be loaded at that point, but that's not guaranteed in any way. This test just makes sure it doesn't crash, if we request the run_info before it was loaded internally. It's a bit hard to explain and may be unnecessary. That's why I made a separate commit for this change; we could just drop it.

M8is: I've added an assert which makes the test a bit more useful. It's essentially checking the default now.

M8is:
Ready for another review.

mubench.pipeline/tests/data/test_detector_run.py
salsolatragus: I'm not sure I grasp the `None` scenario. Where is the `None` coming from? There shouldn't be `None`s flying around anywhere, I think. Is it the test setup that triggers this? But the line 190 should ensure that loading the run info returns an empty dictionary. Can you say a bit more? (Or we can also discuss this on Thursday. It's not critical, I just want to understand).

M8is: When a `DetectorRun` is initialized, we set `self.__run_info = None`. 
This value is only set in two places:
* `__get_run_info(...)`: Load the run info and then acts like a `dict.get(key, default)` and is only called on any direct access on values, e.g. `run.result`.
* `__save_run_info(...)`: Updates run info after write. 

In practice, we always call these at least once before requesting the full run info during publish. The test case was just there to make sure that, if there was any path where none of these were called before, we don't crash. The reason I wasn't sure about that, is this function:

```
    def get_run_info(self):
        run_info = {
            "number_of_findings": self.number_of_findings,
            "runtime": self.runtime,
        }
        run_info.update(self.__run_info)
        return run_info
```

In the scenario of the test case, `self.__run_info` is only accidentally not `None`, because it is loaded on the `self.runtime` call. The test was just there to make sure that this won't break if someone changes the code, as that seems very implicit to me. For example, with the following code, the test will fail:

```
    def get_run_info(self):
        run_info = {
            "number_of_findings": self.number_of_findings,
        }
        run_info.update(self.__run_info)
        return run_info
```

We could probably change `self.__run_info` to a property, which ensures it was loaded before access. This would also make the `self.__get_run_info` obsolete, as we could use `self.__run_info.get` directly. 

Anyway, I thought since it works, a test to make sure it stays working is enough.

salsolatragus: Thanks for the clarification. Now I understand.

Since, apparently, I wasn't aware of this potentially problematic behaviour when I implemented `get_run_info()`, I would prefer to fix this before it actually causes problems. I think using a property like you suggest is the right way to do this. Could you take care of this? You can add it to this PR and I will separate out the two changes.

M8is:
I added the run info refactoring.

Issue #285, opened 2018-02-06
Rename "patterns" to "correct usages"


salsolatragus:
For E1 we currently use the term "pattern" throughout the codebase and in help texts. However, what we talk about are actually not patterns, but correct-usage examples for the detectors to mine patterns from. Since this imprecision in terminology repeatedly confused people, I would like to update it.

I'm aware that there are some locations where this change will break compatibility (e.g., database table names on the review site) and that this change might impact many places.

* [x] Rename folders in data
* [x] Rename Python classes
* [x] Change (log) output

Issue #143, opened 2017-09-13
Rename characteristics to violations


salsolatragus:
Consistently rename characteristics to violations in the dataset, pipeline, and review site.

PR #336, opened 2018-03-27
Now using violations and correct usages (#143 and #285)

M8is:
Implement #143: Rename characteristics to violations
Implement #285: Rename "patterns" to "correct usages"

Some change on the review site may be required, since we changed `patterns` to `correct_usages` in the meta-data upload.


M8is:
@salsolatragus Just pushed the changes you requested.

Issue #321, opened 2018-03-10
Import results from other review site


salsolatragus:
I would like to import results from one review page into another. Ideally, I would like to open a page where I specify a detector and an experiment (and, optionally, a project and version). Then I somehow select the review site to import from (e.g., via supplying the database configuration) and hit a button. As a result, all runs of the specified detector in the specified experiment (for the project/version) are imported (including findings, reviews, tags, etc....).

The import should add the the data of an existing detector with the same muid. In the same way, to reviews should be assigned to existing reviewers with the same name. If a detector or reviewer does not exist, a new one should be created.

PR #338, opened 2018-04-08
Implements #321: Import results from other review site

joschli:


mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -603,4 +635,167 @@ private function deleteOldImages($experimentId, $detectorId, $projectId, $versio
...
+
+                foreach ($misuse->reviews as $review) {
+                    $this->importReviews($review, $cloned_misuse);
+                }
+            }
+        }
+    }
+
+    private function cloneAndResetConnectionModel(Model $model){
+        $clone = $model->replicate();
```
salsolatragus: We have three different terms here: 'clone', 'replicate', and 'import'. I'd appreciate having at most two... replace 'clone'/'cloned' by 'import'/'imported'?

salsolatragus:
Abstract db setup script to allow specifying different connections. This should allow us to write tests from the import.

Issue #325, opened 2018-03-12
Indicate open reviews in menu and on index when logged in


salsolatragus:
When I'm logged in, I would like to see in the menu and on the index page whether there are open reviews/open questions/conflicts for the individual detectors. I imagine that every entry for a detector optionally has icons to its right that indicate this (e.g., the lens for open review).

PR #340, opened 2018-04-08
Implements  #325: Indicate open reviews in menu and on index when logged in

joschli:


mubench.reviewsite/templates/includes/menu-experiment.phtml
```diff
@@ -30,7 +31,11 @@ $is_current_experiment = strpos($path,  "experiments/{$experiment->id}/") !== fa
         <?php foreach ($detectors[$experiment->id] as $detector) : ?>
             <?php $is_current_detector = $this_det && $this_det->muid == $detector->muid ?>
             <li class="<?= $is_current_detector && $is_current_experiment ? "active" : "" ?>">
-                <a href="<?= $pathFor('experiment.detector', array("experiment_id"=>$experiment->id, "detector_muid" => $detector->muid)) ?>"><?= $detector->muid ?></a>
+                <a href="<?= $pathFor('experiment.detector', array("experiment_id"=>$experiment->id, "detector_muid" => $detector->muid)) ?>"><?= $detector->muid ?>
```
salsolatragus: Please right-align the badges.

mubench.reviewsite/templates/index.phtml
```diff
@@ -30,7 +31,11 @@ use MuBench\ReviewSite\Models\Experiment;
                 <ul>
                 <?php if ($detectors[$experiment->id]): ?>
                     <?php foreach ($detectors[$experiment->id] as $detector): ?>
-                        <li><a href="<?= $pathFor('experiment.detector', ['experiment_id' => $experiment->id, 'detector_muid' => $detector->muid]) ?>"><?= $detector->muid ?></a></li>
```
salsolatragus: Please right-align the badges here... I want this to look like a table with a second column showing the icons. Otherwise I image it to be very noisy, if there are many detectors with different names lengths.

mubench.reviewsite/templates/includes/detector_review_state.phtml
```diff
@@ -0,0 +1,20 @@
...
+ * @var array $exp_det_reviewstate contains information about reviewstate ordered by experiment->detector->NEEDS_REVIEW/NEEDS_CLARIFICATION/DISAGGREMENT
+ * @var Experiment $experiment the selected experiment
+ * @var Detector $detector the selected detector
+ */
+
+use MuBench\ReviewSite\Models\Detector;
+use MuBench\ReviewSite\Models\Experiment;
+use MuBench\ReviewSite\Models\ReviewState;
+
+if ($exp_det_reviewstate[$experiment->id][$detector->id][ReviewState::NEEDS_REVIEW]): ?>
```
salsolatragus: Can we somehow distinguish whether I can do another review here or whether all required reviews need to be done by somebody else? I don't habe a really good idea from the top of my head, but, as a reviewer, I would really like to see this.

mubench.reviewsite/bootstrap/renderer.php
```diff
@@ -31,6 +34,31 @@
 
     $markdown_parser = new Parsedown();
     $markdown_parser->setBreaksEnabled(True);
+    $exp_det_reviewstate = [];
```
salsolatragus: This logic should not be here. It should be in some file (class?) dedicated to the menu logic. I imagine that this code here calls a function/creates an object, such that the template can make queries like `$o->hasOpenReviews($detector, $experiment)` to decide what to display.

The logic needs tests and it should be split into multiple functions. There's way to much nesting here.

mubench.reviewsite/templates/includes/detector_review_state.phtml
joschli: The only ideas I have right now are either color coding it or maybe showing the number of reviews you can still review yourself (if any) next to the icon.

salsolatragus: We could change the icon from red to grey if there's no reviews that I can do, yes. That's the best idea so far... I'm trying to think of other systems that have a similar kind of notification badges, but I can't really think of one that shows a badge for both me personally and all users of the system.

Issue #324, opened 2018-03-12
Anonymous mode


salsolatragus:
I want the review site to have an anonymous mode (configurable in the settings) where all reviewer names are displayed as `reviewer-<id>` instead of their true names. Furthermore, in this mode, I want to be able to hide some of the detector's names. For this, I imagine some configuration like this:

    'blind_mode' = [
        'enabled' => true,
        'detector_blind_names' => ['Jadet' => 'Foo']
    ]

joschli:
We talked about using detector ids in the urls in blind mode so the urls don't stop working when enabling/disabling blind mode. What do you think about just translating the blinded detector names to the real detector names independent from the blind mode instead of switching to the ids in the blind mode? This results in the same behaviour but we don't have to use ids in the urls.

salsolatragus:
Functionally, I don't see any real difference between these two alternatives. I'm still in favor of using the Id, because it allows the controller to stay oblivious to the blinding... what's you argument for using the names (besides having consistent URLs between anonymous and public mode, which I don't really count, because no one ever sees both in parallel)?

PR #341, opened 2018-04-12
Implements  #324: Anonymous mode

joschli:


mubench.reviewsite/src/Controllers/Controller.php
```diff
@@ -19,4 +21,13 @@ public function __get($property)
         }
         return null;
     }
+
+    function getDetector($detector_id){
```
salsolatragus: I would prefer to have this on `Detector` as a static method `findByIdOrName()`.

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -28,10 +28,17 @@ public function getReview(Request $request, Response $response, array $args)
         $misuse_muid = $args['misuse_muid'];
 
         $experiment = Experiment::find($experiment_id);
-        $detector = Detector::find($detector_muid);
+        $detector = $this->getDetector($detector_muid);
         $ex2_review_size = $request->getQueryParam("ex2_review_size", $this->settings['default_ex2_review_size']);
 
-        $reviewer = array_key_exists('reviewer_name', $args) ? Reviewer::where(['name' => $args['reviewer_name']])->first() : $this->user;
+        $reviewer = $this->user;
```
salsolatragus: As for the detector, I would move this logic into a lookup method on the `Reviewer` class.

Issue #334, opened 2018-03-22
Integrate cross-project experiment into pipeline


salsolatragus:

M8is:
In [`crossproject_prepare.py`, line 36](https://github.com/stg-tud/MUBench/blob/79455db69a2396518a0709a49479f0aa2af38cb0/mubench.pipeline/crossproject_prepare.py#L36), we open a file for reading for which I can't find where that file would have been written before.  I've searched the repository and this is the only usage I found.
@salsolatragus Do you know how `subtypes.csv` would be created?

salsolatragus:
I created this file manually, but it turned out the experiment didn't actually use it, due do us limiting the number of examples projects to 50 per type. For now, I suggest you assume this file is empty (or non-existent). This is how I used this code in the last submission.

PR #346, opened 2018-04-17
Implement #334: Integrate cross-project experiment into pipeline

M8is:
Implement #334: Integrate cross-project experiment into pipeline

@salsolatragus I thought you may want to check the current state. The tasks are now integrated into the pipeline and should work the same way as before (plus pipeline features like the default config). I've set them up to be called separately, but that is easy to reconfigure.


mubench.pipeline/utils/config_util.py
```diff
@@ -238,6 +249,32 @@ def __add_run_ex3_subprocess(available_detectors: List[str], available_datasets:
     __setup_run_arguments(experiment_parser, available_detectors)
 
 
+def __add_run_cross_project_create_index(available_datasets: List[str], subparsers) -> None:
```
salsolatragus: Implicitly run create index with prepare.

mubench.pipeline/utils/config_util.py
```diff
@@ -238,6 +249,32 @@ def __add_run_ex3_subprocess(available_detectors: List[str], available_datasets:
...
+    parser = subparsers.add_parser("xpci", formatter_class=SortingHelpFormatter,
+                                              help="TODO",
+                                              description="TODO")
+    __setup_filter_arguments(parser, available_datasets)
+
+
+def __add_run_cross_project_prepare(subparsers) -> None:
```
salsolatragus: Provide as top-level task `./mubench checkout-xp`.

salsolatragus:
Try collecting results from AsterskTasks for passing prepared xp data on to subsequent tasks. If this does not work (easily), we persist indexes version-wise and load these after preparation.

M8is:
Turns out we really didn't need anything more since we can write one index per version. Woops.

Well, now that it's implemented we might as well keep it. It isn't that complicated and seems to be a useful feature. Essentially, we can now split something up into arbitrarily small tasks and just accumulate the results. At this point, I'm wondering if `TaskRunner` is Turing complete...

Anyway, we can now run ex2 and ex3 `--with-xp`, which prepares the cross project examples and passes all sources paths to the detector via the training sources argument.

M8is:
@salsolatragus I just remembered that the solution I've implemented now is exactly what you suggested. There's definitely a lot of space for improvements, but it should work for now.

mubench.pipeline/tasks/implementations/detect_all_findings.py
```diff
@@ -20,24 +21,29 @@ def __init__(self, findings_base_path: str, force_detect: bool, timeout: Optiona
...
     @staticmethod
-    def _get_detector_arguments(version_compile: VersionCompile):
+    def _get_detector_arguments(version_compile: VersionCompile, xp_sources_paths: List[str]):
         return {
             key_detector_mode: DetectAllFindingsTask.__DETECTOR_MODE,
             key_target_src_paths: version_compile.original_sources_paths,
             key_target_classes_paths: version_compile.original_classes_paths,
-            key_dependency_classpath: version_compile.get_full_classpath()
+            key_dependency_classpath: version_compile.get_full_classpath(),
+            key_training_src_path: xp_sources_paths
```
salsolatragus: What happens if there is no xp paths.

mubench.pipeline/tasks/implementations/detect_provided_correct_usages.py
```diff
@@ -40,7 +47,7 @@ def _get_findings_path(self, detector: Detector, version: ProjectVersion, misuse
     def _get_detector_arguments(version_compile: VersionCompile, misuse_compile: MisuseCompile):
         return {
             key_detector_mode: DetectProvidedCorrectUsagesTask.__DETECTOR_MODE,
-            key_training_src_path: misuse_compile.correct_usage_sources_path,
+            key_training_src_path: [misuse_compile.correct_usage_sources_path],
```
salsolatragus: Revert this.

M8is:
This should be ready now. 
Note: if you want to use this in the detectors, you'll have to catch the `java.io.FileNotFoundException: no training source path provided`. What is happening is pretty clear from the exception message at least. We'd have to release another `mubench.cli` version to add a default.

Issue #345, opened 2018-04-16
Provide `./mubench update` command


salsolatragus:
This equates to a git pull.

Issue #343, opened 2018-04-13
Add documentation for reviewsite commands


salsolatragus:
[0] $ ./mubench configure review-site
    usage: ./mubench [-h] {check,info,checkout,compile,run,publish,stats} ...
    ./mubench: error: argument task: invalid choice: 'configure' (choose from 'check', 'info', 'checkout', 
    'compile', 'run', 'publish', 'stats')

PR #350, opened 2018-04-20
Refactor mubench CLI (#345 and #343)

M8is:
Implement #343: Add documentation for reviewsite commands
Implement #345: Provide `./mubench update` command

@salsolatragus This is obviously not ready yet, but I'd like to get your opinion on the current idea: 
I've split the `mubench` script up into what I called "meta tasks". This way, we can implement `mubench -h` as an overview of all meta tasks, and `./mubench <meta-task> -h` displays info about the relevant task. Pipeline-specific help can be accessed via `./mubench pipeline -h`.
To maintain compatibility, we have a 'fallthrough' to the pipeline task, meaning that `./mubench <task> ...` will implicitly redirect to `./mubench pipeline <task> ...`, if there is no matching meta task.

My current problem is that we would have to implement something similar for the batch scripts, and I'm not sure how to do that. But maintaining the Windows side is a hassle anyway...


M8is:
> I like the way you solved this. It's elegant and extensible, only it does not really solve the command-line-help problem, does it?

It allows us to have a `mubench -h` that shows all available commands, whereas before we would only show the pipeline help. I didn't implement the actual help yet, but that shouldn't be a problem now. I'll add a little prototype.
Also, we can easily implement separate help files for each command which should be easier to maintain than a single huge documentation for everything.

> I agree that it is a hassle to maintain this for both Linux and Windows, but I guess changes to this should be infrequent and rather small, once we have the infrastructure in place...

Sure. I probably won't be able to get this done before Thursday though.

> Just a thought: If we move the docker command itself to a variable in the env, could we actually write tests for this stuff? Unfortunately, we don't have a Windows CI environment, but if we had something automated, we could at least run it manually on a test machine whenever we make changes...

We could do that. I'm not sure which tool we should use for that. Python? I've never written automated tests for bash scripts.

M8is:
@salsolatragus I've implemented some of the help now.

See [Option A](https://github.com/stg-tud/MUBench/pull/350/commits/07a03d95fabb224d7dba4387fcf13ba134da579f) and [Option B](https://github.com/stg-tud/MUBench/pull/350/commits/5147674cef9927cbba2040f9ed41e501ac42eefc) for the two options of documenting the tasks.


Option A will probably some duplicate code for each task, but has the advantage of giving each task its own responsibility to provide the documentation. This way, the root script only has to handle it's own doc, while the other scripts can essentially do whatever they want.

Option B is less flexible and clutters the root script a bit more but it is very easy to extend. Just add the respective `.txt` file. The implementation right now still allows the task to implement a different approach. This is at the very least necessary for the pipeline, where the documentation comes from python. It may be useful somewhere else though.

M8is:
No idea why that php test fails now.

Help is now implemented as we've discussed, I've also added the python files for `mubench -h` and `mubench browse -h`.

mubench
```diff
@@ -6,16 +6,21 @@ TASKS=${MUBENCH_ROOT}/mubench.bin
...
+  HELP=t
+fi
+
+if [ $HELP ] && [ $# = 1 ]; then
+  docker run --rm -v "$MUBENCH_ROOT":/mubench ${MUBENCH_PIPELINE_DOCKER_IMAGE} python3 ${MUBENCH_DOCS}/mubench.py "$@"
   exit 0
 else
   TASK=${TASKS}/$1
-  DOC="${MUBENCH_DOCS}/$1.txt"
+  DOC=${MUBENCH_DOCS}/$1.py
```
salsolatragus: Why remove the quotes? Does this still work if the path contains a space?

mubench.docs/browse.py
```diff
@@ -0,0 +1,20 @@
+from argparse import HelpFormatter, ArgumentParser
+from operator import attrgetter
+
+import sys
+
+
+class SortingHelpFormatter(HelpFormatter):
```
salsolatragus: I think we don't need the formatter here, because there's no subtasks.

mubench.docs/mubench.py
```diff
@@ -0,0 +1,26 @@
+from argparse import HelpFormatter, ArgumentParser
+from operator import attrgetter
+
+import sys
+
+
+__TASKS = ['pipeline', 'browse', 'reviewsite', 'update']
+
+
+class SortingHelpFormatter(HelpFormatter):
```
salsolatragus: Does this formatter actually sort mandatory arguments? Sorting by `option_strings` appears to me like it shouldn't because there are no option strings for mandatory arguments... (this is not strictly related to this task, but might be a general problem with our CLI help)

salsolatragus:
Seems we have a flaky test (time dependent). You were "just unlucky" with this build. I'll try to fix this as soon as possible.

mubench.docs/mubench.py
M8is: Seems like it doesn't:
```
usage: ./mubench [-h]
                 {check,info,checkout,compile,run,publish,stats,browse} ...

Run MUBench, the benchmark for API-misuse detectors.

positional arguments:
  {check,info,checkout,compile,run,publish,stats,browse}
                        MUBench provides several tasks. Run `./mubench <task>
                        -h` for details.
    check               Check MUBench runtime requirements or dataset
                        consistency.
    info                Show info about projects, project versions, and
                        misuses in MUBench.
    checkout            Checkout project versions.
    compile             Compile project versions. Runs `checkout`, if
                        necessary.
    run                 Run an experiment. Runs `checkout` and `compile`, if
                        necessary.
    publish             Publish data to a review site. Runs `checkout`,
                        `compile`, and `run` , if necessary.
    stats               Print statistics on the MUBench dataset.
    browse              Open a Linux shell in a container mounting the MUBench
                        Docker Volumes.

optional arguments:
  -h, --help            show this help message and exit

For details, check out https://github.com/stg-tud/MUBench.
```

salsolatragus: I created a respective issue #360.

M8is:
Batch files are in place. I've checked if everything runs as expected, but I couldn't test it with docker since I was on Windows home edition. However, the docker commands are called correctly, so I guess it should work.

Issue #51, opened 2017-07-05
Consider code snippets size for chunking requests


mreif:
I tried two upload 200 findings with large Java source files. Experiencing internal server errors, I inspected the code of the publish_finding task. I recognized that the slicing of findings did not work in my case. Every finding had exactly one file which was not found in the loop of publish_findings_task.py => _slice_by_max_files_per_post.

salsolatragus:
If I get your setting correctly, then the problem is that the snippets are not send as files, but as strings, and are, thus, not considered for the slicing. We currently slice only because PHP doesn't accept sending more than 20 files in one request (by the default setting).

Can you elaborate more on the internal server error that you faced? Is it that the total request size gets too large for the server? In that case we should add an additional slicing criterion, I guess...

mreif:
Sorry, I didn't notice that commented the issue... Whatever, I assumed that the snippets are sent as files. Since they are not, your assumption is correct, the total request size got too large and as result, everything crashed. I think it would make sense to either send source snippets as files (this might be too much work) or add an additional slicing criterion that also considers other parts of the request.

salsolatragus:
Don't worry, I had plenty of other things to do ;)

Since our usual scenario is to extract only methods (as opposed to full files, as in your case), sending files will requires to create new files from the strings we have in memory, just for sending. I'd prefer to avoid that. However, it should be possible to determine the size of the request sent to the server from the size of the data (including the snippets) and to use that as an additional slicing criterium. We'll look into it. Thanks for clarifying!

PR #351, opened 2018-04-20
Implement #51: Consider code snippets size for chunking requests

M8is:
Implement #51: Consider code snippets size for chunking requests

We now additionally limit the slices by potential-hit size in bytes. @salsolatragus I've set a somewhat arbitrary limit, so you'll probably want to adjust that.

### DIFF
--- a/mubench.pipeline/tasks/implementations/publish_findings.py
+++ b/mubench.pipeline/tasks/implementations/publish_findings.py
@@ -1,12 +1,10 @@
 import getpass
 import logging
-from sys import getsizeof
+import re
+from os.path import getsize
 from typing import List, Dict
 from urllib.parse import urljoin
 
-import re
-
-from os.path import getsize
 from requests import RequestException
 
 from data.detector import Detector
@@ -83,7 +81,7 @@ class PublishFindingsTask:
         size_of_slice = 0
         for potential_hit in potential_hits:
             number_of_files_in_hit = len(potential_hit.files)
-            size_of_hit = self.__get_potential_hit_size(potential_hit)
+            size_of_hit = total_size(potential_hit)
             exceeds_max_files_per_post = number_of_files_in_slice + number_of_files_in_hit > self.max_files_per_post
             exceeds_max_post_size = size_of_slice + size_of_hit > self.max_post_size_in_bytes
             if potential_hits_slice and (exceeds_max_files_per_post or exceeds_max_post_size):
@@ -160,10 +158,6 @@ class PublishFindingsTask:
 
         return files
 
-    @staticmethod
-    def __get_potential_hit_size(potential_hit: 'SpecializedFinding') -> int:
-        return getsizeof(potential_hit) + sum([getsize(file) for file in potential_hit.files])
-
 
 class SpecializedFinding(Finding):
     def __init__(self, data: Dict[str, str], files: List[str] = None):
@@ -171,4 +165,4 @@ class SpecializedFinding(Finding):
         self.files = files or []
 
     def __sizeof__(self):
-        return total_size(self.__dict__) + total_size(self.files)
+        return total_size(self.__dict__) + sum([getsize(file) for file in self.files])
--- a/mubench.pipeline/tasks/implementations/publish_findings.py
+++ b/mubench.pipeline/tasks/implementations/publish_findings.py
@@ -26,7 +26,7 @@ class PublishFindingsTask:
                  review_site_password: str = ""):
         super().__init__()
         self.max_files_per_post = 20  # 20 is PHP's default limit to the number of files per request
-        self.max_file_size_per_post = 5000  # choose a moderate value since we only approximate upload size
+        self.max_post_size_in_bytes = 5000  # choose a moderate value since we only approximate upload size
 
         self.experiment_id = experiment_id
         self.compiles_base_path = compiles_base_path
@@ -63,7 +63,7 @@ class PublishFindingsTask:
             for potential_hit in potential_hits.findings]
 
         try:
-            for postable_potential_hits_slice in self.__slice_by_max_files_and_size_per_post(postable_potential_hits):
+            for postable_potential_hits_slice in self.__slice_by_number_of_files_and_post_size(postable_potential_hits):
                 file_paths = self.__get_file_paths(postable_potential_hits_slice)
                 postable_data = self.__to_postable_data(run_info, result, postable_potential_hits_slice)
                 self.__post(project, version, detector, postable_data, file_paths)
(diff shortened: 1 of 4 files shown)


mubench.pipeline/tests/tasks/implementations/test_publish_findings_task.py
```diff
@@ -49,15 +50,17 @@ def setup(self):
 
         self.uut = PublishFindingsTask(self.experiment_id, "/sources", "http://dummy.url", "-username-", "-password-")
 
-    def test_post_url(self, post_mock):
+    def test_post_url(self, post_mock, getsize_mock):
+        getsize_mock.return_value = 0
```
salsolatragus: Do we really need to do this in every test? Doesn't the patch use some sensible default? Or can it not, due to no type information? If so, can we move this to the setup somehow?

mubench.pipeline/tests/tasks/implementations/test_publish_findings_task.py
```diff
@@ -273,6 +289,24 @@ def test_converts_directed_graphs(self, replace_mock, post_mock):
 
         assert_equals([self.test_detector_execution.findings_path + "/graph"], post_mock.call_args[1]["file_paths"])
 
+    @patch("tasks.implementations.publish_findings.PublishFindingsTask._convert_graphs_to_files")
+    def test_publish_successful_run_in_sized_chunks(self, convert_mock, post_mock, getsize_mock):
+        self.uut.max_files_per_post = 100
```
salsolatragus: Why do we need to set this here? It doesn't seem to be relevant for the test.

mubench.pipeline/tests/tasks/implementations/test_publish_findings_task.py
```diff
@@ -273,6 +289,24 @@ def test_converts_directed_graphs(self, replace_mock, post_mock):
...
+            self._create_finding({"rank": "-1-"}, convert_mock, file_paths=["-file1-"]),
+            self._create_finding({"rank": "-2-"}, convert_mock, file_paths=["-file2-"])
+        ])
+
+        self.uut.run(self.project, self.version, self.test_detector_execution, self.test_potential_hits,
+                     self.version_compile, self.detector)
+
+        assert_equals(len(post_mock.call_args_list), 2)
+        assert_equals(["-file1-"], post_mock.call_args_list[0][1]["file_paths"])
+        assert_equals(["-file2-"], post_mock.call_args_list[1][1]["file_paths"])
```
salsolatragus: Do we even need files in this test? Its name says it's supposed to only test the chunking by size, so I would remove the files from the findings and the last two assertions.

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -25,6 +26,7 @@ def __init__(self, experiment_id: str, compiles_base_path: str, review_site_url:
                  review_site_password: str = ""):
         super().__init__()
         self.max_files_per_post = 20  # 20 is PHP's default limit to the number of files per request
+        self.max_file_size_per_post = 5000  # choose a moderate value since we only approximate upload size
```
salsolatragus: `max_post_size_in_bytes`?

PHP defaults to 2MB per file and 8MB for the whole request. I would probably set 7MB here, to be on the safe side.

Looking at this, we should consider the size of the image files as well... can we do that?

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -72,18 +74,24 @@ def run(self, project: Project, version: ProjectVersion, detector_run: DetectorR
             else:
                 logger.error("%s", e)
 
-    def __slice_by_max_files_per_post(self, potential_hits: List['SpecializedFinding']) -> List[List[Finding]]:
+    def __slice_by_max_files_and_size_per_post(self, potential_hits: List['SpecializedFinding']) -> List[List[Finding]]:
```
salsolatragus: `_slice_by_number_of_files_and_post_size`?

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -72,18 +74,24 @@ def run(self, project: Project, version: ProjectVersion, detector_run: DetectorR
...
         potential_hits_slice = []
         number_of_files_in_slice = 0
+        size_of_slice = 0
         for potential_hit in potential_hits:
             number_of_files_in_hit = len(potential_hit.files)
-            if number_of_files_in_slice + number_of_files_in_hit > self.max_files_per_post:
+            size_of_hit = self.__get_potential_hit_size(potential_hit)
+            file_count_limit = number_of_files_in_slice + number_of_files_in_hit > self.max_files_per_post
```
salsolatragus: `exceeds_max_files_per_post`?

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -72,18 +74,24 @@ def run(self, project: Project, version: ProjectVersion, detector_run: DetectorR
...
+            file_size_limit = size_of_slice + size_of_hit > self.max_file_size_per_post
```
salsolatragus: `exceeds_max_post_size`?

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -72,18 +74,24 @@ def run(self, project: Project, version: ProjectVersion, detector_run: DetectorR
...
+            if potential_hits_slice and (file_count_limit or file_size_limit):
```
salsolatragus: Was there a particular reason you changed this condition to ensure there's always at least one potential hit in a slice? I like this in general, just curious... Do we have a test that checks we include at least one potential hit, even if both the max size and the max number of files are exceeded?

salsolatragus:
The Python doc says that [getsizeof](https://docs.python.org/3/library/sys.html#sys.getsizeof) is non-recursive. Does that impact our computation here? (There's a mention of an alternative recursive recipe)

mubench.pipeline/tasks/implementations/publish_findings.py
M8is: I guess it's not necessary to do this. After the `yield`, we add the potential hit to the next slice anyway. When I added this, I thought we needed this to ensure that we don't silently drop any potential hits. 
Still, I think it makes sense to ensure that we don't return an empty slice. I'll write a test for it.

M8is:
Ready for another review.

mubench.pipeline/tasks/implementations/publish_findings.py
```diff
@@ -149,8 +160,15 @@ def _convert_graphs_to_files(potential_hit: Dict, findings_path: str) -> List[st
 
         return files
 
+    @staticmethod
+    def __get_potential_hit_size(potential_hit: 'SpecializedFinding') -> int:
+        return getsizeof(potential_hit) + sum([getsize(file) for file in potential_hit.files])
```
salsolatragus: Just for me to understand how this works: `getsizeof` calls `__sizeof__`, right? So the call above invokes the method declared below. If this is the case, we add the size of the dictionary (key + values), the size of the file list (i.e., the file names), and the size of the files themselves. To make this more obvious, could we move the `sum` to the `__sizeof__` method and replace the `__get_potential_hit_size` by `total_size(potential_hit)`?

mubench.pipeline/utils/size.py
```diff
@@ -0,0 +1,44 @@
...
+    Automatically finds the contents of the following builtin containers and
+    their subclasses:  tuple, list, deque, dict, set and frozenset.
+    To search other containers, add handlers to iterate over their contents:
+
+        handlers = {SomeContainerClass: iter,
+                    OtherContainerClass: OtherContainerClass.get_elements}
+
+    """
+    seen = set()  # track which object id's have already been seen
+    default_size = getsizeof(0)  # estimate sizeof object without __sizeof__
```
salsolatragus: Do I get this right that this calculates the size of *an* object (0, in this case) as an estimate of the minimal memory footprint of any object? I was confused that this does not use `o` at first.

mubench.pipeline/utils/size.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def total_size(o, verbose=False):
+    """ Returns the approximate memory footprint an object and all of its contents.
+
+    Automatically finds the contents of the following builtin containers and
+    their subclasses:  tuple, list, deque, dict, set and frozenset.
+    To search other containers, add handlers to iterate over their contents:
+
+        handlers = {SomeContainerClass: iter,
```
salsolatragus: Since `__handlers` is private, how can I, as a client, add handlers?

M8is:
Ready for another review.

PR #361, opened 2018-05-03
Resolve #327: Modify tags without reloading the page

### LINKED ISSUE #327: Modify tags without reloading the page
I would like that adding/removing tags to a misuse does not reload the page, but does an asynchronous request and updates the list of tags in place. In fact, I like this to happen on some more places on the site, but I think the tags are a good place to try this out.

joschli:


mubench.reviewsite/templates/review.phtml
```diff
@@ -432,7 +426,30 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
...
+                            'misuse_muid' => $misuse->misuse_muid)) ?>";
+                        var originUrl = "<?= htmlspecialchars($path . "?origin=" . $origin_path); ?>";
+                        var baseRemoveUrl = "<?= $pathFor('tag.remove', array('experiment_id' => $experiment->id,
+                            'detector_muid' => $detectorPathId($detector),
+                            'project_muid' => $misuse->getProject(),
+                            'version_muid' => $misuse->getVersion(),
+                            'misuse_muid' => $misuse->misuse_muid,
+                            'tag_id' => 0)) ?>";
+                        // remove '0/delete' from baseRemoveUrl
+                        baseRemoveUrl = baseRemoveUrl.substring(0, baseRemoveUrl.length - 8);
```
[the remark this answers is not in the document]
salsolatragus: Why are we generating this URL here, still? Shouldn't this be gone now that we generate and return the URL in the controller? The URL is generated for dynamically created tags in the controller and for pre-existing in generating the delete buttons above (line 111). What am I missing?

mubench.reviewsite/templates/review.phtml
```diff
@@ -102,27 +102,26 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
...
                             </div>
                         <?php endforeach; ?>
                     <?php endif; ?>
+                    </div>
                     <?php if($user) : ?>
                             <div style="display:inline-block;height:100%;padding:4px;">
-                                <form id="tag-form" action="<?= $pathFor('tag.add', array('experiment_id' => $experiment->id, 'detector_muid' => $detectorPathId($detector), 'project_muid' => $misuse->getProject(), 'version_muid' => $misuse->getVersion(), 'misuse_muid' => $misuse->misuse_muid)) ?>" method="post" target="">
-                                    <input type="hidden" name="misuse_id" value="<?= htmlspecialchars($misuse->id); ?>">
-                                    <input type="hidden" name="path" value="<?= htmlspecialchars($path . "?origin=" . $origin_path); ?>">
+                                    <input type="hidden" name="misuse_id" value="">
```
salsolatragus: What is this good for? There's no form submitted anymore.

mubench.reviewsite/js/tag-util.js
```diff
@@ -0,0 +1,47 @@
+function addTag(event){
+    if (event.key === "Enter") {
+        let misuseTagsDiv = document.getElementById('misuse-tags');
+        tagName = event.target.value;
+        if (misuseTagsDiv.querySelectorAll("span#" + tagName).length == 0) {
+            let msg = {
+                "tag_name": tagName,
+                "misuse_id": misuseId
+            };
+            postJson(addUrl, msg, function(r){
```
salsolatragus: The `addUrl` should be a parameter to this function and not a globally defined variable that is assumed to exist. For remove you already did this.

mubench.reviewsite/js/tag-util.js
```diff
@@ -0,0 +1,47 @@
+function addTag(event){
+    if (event.key === "Enter") {
+        let misuseTagsDiv = document.getElementById('misuse-tags');
+        tagName = event.target.value;
+        if (misuseTagsDiv.querySelectorAll("span#" + tagName).length == 0) {
+            let msg = {
```
salsolatragus: Rename `msg` to `data`?

mubench.reviewsite/js/tag-util.js
```diff
@@ -0,0 +1,47 @@
+function addTag(event){
+    if (event.key === "Enter") {
+        let misuseTagsDiv = document.getElementById('misuse-tags');
+        tagName = event.target.value;
+        if (misuseTagsDiv.querySelectorAll("span#" + tagName).length == 0) {
```
salsolatragus: Extract this into a helper named `isExistingTag`?

mubench.reviewsite/js/tag-util.js
```diff
@@ -0,0 +1,47 @@
+function addTag(event){
+    if (event.key === "Enter") {
```
salsolatragus: I'm wondering whether this event handling should be part of the handler (`onKeyUp="if (event.key == 'Enter') addTag(url, this.value)"`) and this function receive the new tag name instead of the event? I'm really not familiar with the Javascript best practices here, but this seems to mix so many different levels of responsibility (event handling,  dom handling, sending requests)...

mubench.reviewsite/templates/review.phtml
```diff
@@ -432,7 +431,30 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
                     simplemde.codemirror.refresh();
                     document.getElementById("comment-area").style.display="none";
                     <?php endif; ?>
+                    <?php if($user): ?>
+                        var addUrl = "<?= $pathFor('tag.add', array('experiment_id' => $experiment->id,
```
salsolatragus: I don't like that this is in a dedicated script while the setup for removing tags is inline in the button event handler. I think it should be consistent for both cases. Since its not too much code, I would probably inline it, but I don't really have a strong preference here.

PR #365, opened 2018-05-13
Resolve #348: Cleanup creation of database tables

### LINKED ISSUE #348: Cleanup creation of database tables
Simplify script with a helper that takes care of setting the right connection and of dropping existing tables.

joschli:


mubench.reviewsite/setup/TableCreationHelper.php
```diff
@@ -0,0 +1,29 @@
...
+
+    /** @var \Illuminate\Database\Schema\Builder */
+    protected $schema;
+
+    function __construct($connection)
+    {
+        $this->schema = Schema::connection($connection);
+    }
+
+    function dropIfExistsAndCreateNewWith($modelClass, $createFunc)
```
salsolatragus: I'm not happy with the name of this method. I would like it to have the same name as the method below, which, in PHP, probably requires us to merge this into one method that type-checks the first parameter (Model vs. str).

mubench.reviewsite/setup/db_table_utils.php
```diff
@@ -0,0 +1,159 @@
+<?php
+use Illuminate\Database\Schema\Blueprint;
+
+require_once __DIR__ . "/TableCreationHelper.php";
+
+function createTables($connection)
```
salsolatragus: I like that this is a function now that we can call with the connection. I'm wondering whether we want to go one step further an make this a method on the `TableCreationHelper` (which should then be called something like `SchemaSetup`). Then we could instantiate a setup for a connection and just `run()` it. What do you think?

mubench.reviewsite/setup/schema_util.php
```diff
@@ -0,0 +1,162 @@
+<?php
+
+use Illuminate\Database\Schema\Blueprint;
+
+require_once __DIR__ . "/Schema.php";
```
salsolatragus: Can't we replace this by a `use` statement?

mubench.reviewsite/setup/schema_util.php
```diff
@@ -0,0 +1,162 @@
+<?php
```
salsolatragus: I think I would prefer a filename like `database_setup_utils`... not perfect, but it's not a schema utils anymore.

mubench.reviewsite/tests/ImportRunsControllerTest.php
```diff
@@ -164,10 +164,9 @@ private function addDBConnections($firstdb, $seconddb)
 
     private function setupDatabaseTables()
     {
-        $schema = Schema::connection('default');
-        require __DIR__ . '/../setup/create_database_tables.php';
-        $schema = Schema::connection('extern');
-        require __DIR__ . '/../setup/create_database_tables.php';
+        foreach(array('default', 'extern') as $connection){
+            createTables($connection);
+        }
```
salsolatragus: How about

    createTables('default');
    createTables('extern');

? xD

PR #366, opened 2018-05-13
Resolve  #273: Hide tags on review page before agreement

### LINKED ISSUE #273: Hide tags on review page before agreement
Tags should not be visible on the review page until agreement is reached. This includes that reviewers are unable to add tags before that point. Although I would like to do this during reviews, I don't want others to see my tags. I think other solutions (e.g., storing who added a tag and displaying selectively) to this problem are way more complex.

joschli:
I added a pivot column for the reviewer id to the tags. This enables us to show personal tags only on the review page when the review isn't conclusive yet. On the detector page we already show tags only when we have a conclusive review decision for a misuse. Should we change this to personal tags as well until we have a conclusive review or keep the current state?

### DIFF
--- a/mubench.reviewsite/src/Controllers/ReviewsController.php
+++ b/mubench.reviewsite/src/Controllers/ReviewsController.php
@@ -148,7 +148,7 @@ class ReviewsController extends Controller
             $findingReview = FindingReview::firstOrNew(['review_id' => $review->id, 'rank' => $rank]);
             $findingReview->decision = $findings_review['hit'];
             $findingReview->save();
-            $findingReview->violations()->sync(array_key_exists('violations', $findings_review) ? $findings_review['violations'] : array());
+            $findingReview->violations()->sync($findings_review['violations']);
         }
     }
 
--- a/mubench.reviewsite/src/Controllers/ReviewsController.php
+++ b/mubench.reviewsite/src/Controllers/ReviewsController.php
@@ -119,10 +119,11 @@ class ReviewsController extends Controller
         $comment = $review['review_comment'];
         $misuse_id = $review['misuse_id'];
         $hits = $review['review_hit'];
+        $tags = array_key_exists('review_tags', $review) ? $review['review_tags'] : array();
 
         $reviewer_name = $args['reviewer_name'];
         $reviewer = Reviewer::findByIdOrName($reviewer_name);
-        $this->updateOrCreateReview($misuse_id, $reviewer->id, $comment, $hits);
+        $this->updateOrCreateReview($misuse_id, $reviewer->id, $comment, $hits, $tags);
 
         if ($review["origin"] != "") {
             return $response->withRedirect("{$review["origin"]}");
@@ -137,17 +138,17 @@ class ReviewsController extends Controller
         }
     }
 
-    public function updateOrCreateReview($misuse_id, $reviewer_id, $comment, $findings_reviews_by_rank)
+    public function updateOrCreateReview($misuse_id, $reviewer_id, $comment, $findings_reviews_by_rank, $tags)
     {
         $review = Review::firstOrNew(['misuse_id' => $misuse_id, 'reviewer_id' => $reviewer_id]);
         $review->comment = $comment;
         $review->save();
-
+        TagsController::syncReviewTags($review->id, $tags);
         foreach ($findings_reviews_by_rank as $rank => $findings_review) {
             $findingReview = FindingReview::firstOrNew(['review_id' => $review->id, 'rank' => $rank]);
             $findingReview->decision = $findings_review['hit'];
             $findingReview->save();
-            $findingReview->violations()->sync($findings_review['violations']);
+            $findingReview->violations()->sync(array_key_exists('violations', $findings_review) ? $findings_review['violations'] : array());
         }
     }
 
--- a/mubench.reviewsite/src/Controllers/TagsController.php
+++ b/mubench.reviewsite/src/Controllers/TagsController.php
@@ -36,7 +36,6 @@ class TagsController extends Controller
 
     public function getTags(Request $request, Response $response, array $args)
     {
-        // TODO: fix tag statistics with reviews
         $tags = Tag::all();
         $results = array(1 => array(), 2 => array(), 3 => array());
         $totals = array(1 => array(), 2 => array(), 3 => array());
--- a/mubench.reviewsite/src/Controllers/TagsController.php
+++ b/mubench.reviewsite/src/Controllers/TagsController.php
@@ -41,10 +41,11 @@ class TagsController extends Controller
         $results = array(1 => array(), 2 => array(), 3 => array());
         $totals = array(1 => array(), 2 => array(), 3 => array());
         foreach($tags as $tag){
-            $tagged_misuses = $tag->misuses;
-            foreach($tagged_misuses as $misuse){
-                $results[$misuse->run->experiment_id][$misuse->detector->muid][$tag->name][] = $misuse;
-                $totals[$misuse->run->experiment_id][$tag->name][] = $misuse;
+            /** @var Review $review */
+            foreach($tag->reviews as $review){
+                $misuse = $review->misuse;
+                $results[$misuse->run->experiment_id][$misuse->detector->muid][$tag->name][$misuse->id] = $misuse;
+                $totals[$misuse->run->experiment_id][$tag->name][$misuse->id] = $misuse;
             }
         }
         foreach($totals as $key => $total){
--- a/mubench.reviewsite/src/Controllers/TagsController.php
+++ b/mubench.reviewsite/src/Controllers/TagsController.php
@@ -4,29 +4,13 @@ namespace MuBench\ReviewSite\Controllers;
 
 
 use MuBench\ReviewSite\Models\Misuse;
+use MuBench\ReviewSite\Models\Review;
 use MuBench\ReviewSite\Models\Tag;
 use Slim\Http\Request;
 use Slim\Http\Response;
 
 class TagsController extends Controller
 {
-    public function postTag(Request $request, Response $response, array $args)
-    {
-        $formData = $request->getParsedBody();
-        $tag_id = $formData['tag_name'];
-        $misuse_id = $formData['misuse_id'];
-        $reviewer_id = $formData['reviewer_id'];
-
-        $tag = $this->addTagToMisuse($misuse_id, $tag_id, $reviewer_id);
-        $removeUrl = $this->router->pathFor('private.tag.remove', array('experiment_id' => $args["experiment_id"],
-            'detector_muid' => $args["detector_muid"],
-            'project_muid' => $args["project_muid"],
-            'version_muid' => $args["version_muid"],
-            'misuse_muid' => $args["misuse_muid"],
-            'tag_id' => $tag->id));
-        
-        $response->withJson(array("id" => $tag->id, "color" => $tag->color, "fontColor" => $tag->getFontColor(), "removeUrl" => $removeUrl));
-    }
 
     public function manageTags(Request $request, Response $response, array $args)
     {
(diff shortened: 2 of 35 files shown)


mubench.reviewsite/templates/detector.phtml
```diff
@@ -255,7 +255,7 @@ $columns_to_ignore = array(
                 <td class="<?= $style_classes ?>">
                     <?php if($misuse->hasConclusiveReviewState()) : ?>
                         <?php if($misuse->misuse_tags): ?>
-                            <?php foreach($misuse->misuse_tags as $tag) : ?>
+                            <?php foreach($misuse->misuse_tags->unique('id') as $tag) : ?>
```
salsolatragus: We should use `getTags()` here and remove the surrounding check for `hasConclusiveReviewState()`. Then i consistently see my own tags during review and all tags after reviews reached a conclusive state.

joschli:
I realised that you could see another reviewers tags when you visit his review. When a review is conclusive we are showing all tags on all the reviews. How should we handle this? What do you want to see when visiting another review?

salsolatragus:
It's okay to show the other reviewer's tags when viewing his review, because then I see his decision anyways. We're not trying to prevent reviewers from seeing what other did. The goal is that a reviewer who does not want to peek, doesn't do so accidentally.

mubench.reviewsite/templates/review.phtml
```diff
@@ -106,11 +106,11 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
             <?php if($experiment->id !== 2) : ?>
                 <tr><td><b>Misuse:</b></td><td>misuse '<?= $markdown_parser->text($misuse->misuse_muid); ?>'</td></tr>
             <?php endif; ?>
-            <?php if($user || $misuse->misuse_tags) : ?>
```
salsolatragus: Check if we can make the property private.

mubench.reviewsite/tests/create_simple_test_data.php
```diff
@@ -448,6 +448,6 @@
 $tag->save();
 
 echo 'Creating MisuseTags<br/>';
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 1, 'reviewer_id' => $reviewer->id));
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 1, 'tag_id' => 2, 'reviewer_id' => $reviewer->id));
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 2, 'reviewer_id' => $reviewer->id));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 1, 'reviewer_id' => 2));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 1, 'tag_id' => 2, 'reviewer_id' => 3));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 2, 'reviewer_id' => 3));
```
salsolatragus: Please revert to use actual reviewers again to establish the references.

mubench.reviewsite/src/Models/Misuse.php
```diff
@@ -229,4 +229,19 @@ public function hasReviewed($reviewer)
         return false;
     }
 
+    public function getTags($reviewer)
```
salsolatragus: Ensure tests.

mubench.reviewsite/tests/create_simple_test_data.php
```diff
@@ -448,6 +448,6 @@
 $tag->save();
 
 echo 'Creating MisuseTags<br/>';
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 1));
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 1, 'tag_id' => 2));
-$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 2));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 1, 'reviewer_id' => 2));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 1, 'tag_id' => 2, 'reviewer_id' => 3));
+$capsule->table('misuse_tags')->insert(array('misuse_id' => 3, 'tag_id' => 2, 'reviewer_id' => 3));
```
salsolatragus: Please use reviewers to establish the references (instead of hardcoded ids).

joschli:
I found another Problem with making misuse_tags private. We also need it when importing a run if we want to keep what tags a specific reviewer added. We could bypass this by making a custom query, but I think keeping misuse_tags public would be the cleaner solution.

salsolatragus:
This statement confuses me. Doesn't `misuse_tags` return the tags assigned by any reviewer? If so, how does it help us identify which tags a specific reviewer added? Isn't the problem rather that, if we hide the attribute, we have to query `getTags($reviewer)` for all reviewers in order to migrate?

Risking your eternal hate:

Isn't this whole problem due to our modelling? The tags are now essentially a property of the review, if we would put the attribute on `Review`, this would always give us the tags assigned by the respective reviewer. We could then accumulate all tags in the misuse by either querying the current reviewer's tags or accumulating all reviewers' tags (depending on the review state).

The problem this poses to our current solution is that we then need to create a review to store tags. Then it would make sense to make tags a part of the review in the UI as well and save them only with the review, wouldn't it? 🙈😇

joschli:
The problem with just querying getTags with all reviewers is that we lose some information when the review is already conclusive.

To your second point: I had the exact same thought. When I was testing this feature it bugged me that you could add tags to a misuse without making a review even though they are reviewer related now. 
It makes sense to link them directly to reviews.

salsolatragus:
Ok, let's move tags to reviews.

joschli:
The tags are now connected to a review and you can add and remove them which is only saved if you actually submit a review. I haven't moved the tag box to the comment area yet.
Do we count tags on misuses that don't have a conclusive review in the tag statistic ?

salsolatragus:
Yes, please do count them. I don't see any reason for making a distinction at this point.

mubench.reviewsite/src/Models/Review.php
```diff
@@ -26,6 +26,11 @@ public function tags()
         return $this->belongsToMany(Tag::class, 'review_tags', 'review_id', 'tag_id');
     }
 
+    public function misuse()
+    {
+        return $this->belongsTo(Misuse::class);
```
salsolatragus: Does this use the 'misuse_id' field automatically? If it doesn't we should ensure it does, to prevent managing the same link twice!

mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -137,17 +138,17 @@ public function postReview(Request $request, Response $response, array $args)
...
         $review->comment = $comment;
         $review->save();
-
+        TagsController::syncReviewTags($review->id, $tags);
         foreach ($findings_reviews_by_rank as $rank => $findings_review) {
             $findingReview = FindingReview::firstOrNew(['review_id' => $review->id, 'rank' => $rank]);
             $findingReview->decision = $findings_review['hit'];
             $findingReview->save();
-            $findingReview->violations()->sync($findings_review['violations']);
+            $findingReview->violations()->sync(array_key_exists('violations', $findings_review) ? $findings_review['violations'] : array());
```
salsolatragus: This seems unrelated to tags?

mubench.reviewsite/src/Controllers/TagsController.php
```diff
@@ -42,39 +27,25 @@ public function updateTags(Request $request, Response $response, array $args)
...
-        $formData = $request->getParsedBody();
-        $tag_id = $args['tag_id'];
-        $misuse_id = $formData['misuse_id'];
-
-        $this->deleteTagFromMisuse($misuse_id, $tag_id);
-    }
-
     public function getTags(Request $request, Response $response, array $args)
     {
+        // TODO: fix tag statistics with reviews
```
salsolatragus: Remove now?

mubench.reviewsite/templates/review.phtml
```diff
@@ -106,34 +106,26 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
...
                                 <span id="<?= htmlspecialchars($tag->name) ?>"><?= htmlspecialchars($tag->name) ?></span>
                                 <?php if($user) : ?>
-                                    <button onclick="removeTag('<?= $pathFor('tag.remove', array('experiment_id' => $experiment->id,
-                                        'detector_muid' => $detectorPathId($detector),
-                                        'project_muid' => $misuse->getProject(),
-                                        'version_muid' => $misuse->getVersion(),
-                                        'misuse_muid' => $misuse->misuse_muid,
-                                        'tag_id' => $tag->id)) ?>', '<?= $misuse->id ?>', this.parentElement)"style="border:none;outline:none;background-color:<?= htmlspecialchars($tag->color) ?>;margin:0;padding:0"><i class="fa fa-trash"></i></button>
+                                    <button onclick="removeTag(this.parentElement)"style="border:none;outline:none;background-color:<?= htmlspecialchars($tag->color) ?>;margin:0;padding:0"><i class="fa fa-trash"></i></button>
```
salsolatragus: This is the tag display at the top of the page, right? What happens if I click remove here? The tag vanishes, but since this is not immediately synced to the server, it gets lost with the next action in any case, doesn't it? Shouldn't this be display-only?

mubench.reviewsite/templates/review.phtml
```diff
@@ -371,6 +363,26 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
                             <div class="comment-overlay-button" id="overlay-button" style="z-index:3"><span>This misuse was already reviewed.</span><a class="button" onclick="review(this)">review it anyways</a></div>
                         <?php endif; ?>
                         <table class="invisible">
+                            <tr><td>
+                                    <table class="invisible">
+                                        <tr><td>Tags:</td><td><div class="misuse-tag" style="background-color:<?= htmlspecialchars($tag->color) ?>;color:<?= htmlspecialchars($tag->getFontColor()) ?>;">
+                                                    <span id="<?= htmlspecialchars("Test") ?>"><?= htmlspecialchars("Test") ?></span>
+                                                    <?php if($user) : ?>
```
salsolatragus: Doesn't `$is_reviewer && $reviewer` above imply `$user`?

mubench.reviewsite/templates/review.phtml
```diff
@@ -371,6 +363,26 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
                             <div class="comment-overlay-button" id="overlay-button" style="z-index:3"><span>This misuse was already reviewed.</span><a class="button" onclick="review(this)">review it anyways</a></div>
                         <?php endif; ?>
                         <table class="invisible">
+                            <tr><td>
+                                    <table class="invisible">
+                                        <tr><td>Tags:</td><td><div class="misuse-tag" style="background-color:<?= htmlspecialchars($tag->color) ?>;color:<?= htmlspecialchars($tag->getFontColor()) ?>;">
+                                                    <span id="<?= htmlspecialchars("Test") ?>"><?= htmlspecialchars("Test") ?></span>
```
salsolatragus: Hardcoded "Test" seems strange...

mubench.reviewsite/templates/review.phtml
```diff
@@ -371,6 +363,26 @@ $origin_path = $origin_path ? $origin_path : $pathFor('experiment.detector', arr
...
+                                                    <?php if($user) : ?>
+                                                        <button onclick="removeTag(this.parentElement)"style="border:none;outline:none;background-color:<?= htmlspecialchars($tag->color) ?>;margin:0;padding:0"><i class="fa fa-trash"></i></button>
+                                                    <?php endif; ?>
+                                                </div>
+                                                <?php if($user) : ?>
```
salsolatragus: Same regarding `$user` here.

mubench.reviewsite/tests/SlimTestCase.php
```diff
@@ -177,4 +177,10 @@ protected function createFindingWith($experiment, $detector, $misuse)
         $finding->method = "method(A)";
         $finding->save();
     }
+
+    function createReview($misuse, $reviewer, $hit, $violations = [], $tags = [])
```
salsolatragus: This does not belong into this base class. Since it's reused between many tests, may place it in a `review_test_utils` or something?

PR #374, opened 2018-05-15
Resolve #352: Select smaller runner interface version

### LINKED ISSUE #352: Select smaller runner interface version
We've repeatedly released new versions with changes only on the Java side, e.g., I just released 0.0.13. If I now update the detector releases and specify 0.0.13, the pipeline needs a 0.0.13 runner interface or this will crash. This means I have to create a new Python runner interface, even if I only change the Java side (this interface is basically an empty class, which contains nothing but the version number).

To make my life easier, I would like the logic to automatically select the next lower version number, in case there is not runner interface for the exact same version number. This should behave exactly like we want, but saves us all these empty interfaces.

M8is:
Resolve #352: Select smaller runner interface version

### DIFF
--- a/mubench.pipeline/data/runner_interface.py
+++ b/mubench.pipeline/data/runner_interface.py
@@ -206,18 +206,7 @@ class RunnerInterface_0_0_8(CommandLineArgsRunnerInterface):
 
 
 # noinspection PyPep8Naming
-class RunnerInterface_0_0_10(RunnerInterface_0_0_8):
-    @staticmethod
-    def version():
-        return StrictVersion("0.0.10")
-
-    @staticmethod
-    def changelog():
-        return ""  # Only changed Java-side interface.
-
-
-# noinspection PyPep8Naming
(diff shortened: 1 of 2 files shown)


mubench.pipeline/data/runner_interface.py
```diff
@@ -37,11 +32,9 @@ def __init__(self, jar_path: str, java_options: List[str]):
     @staticmethod
     def get(requested_version: StrictVersion, jar_path: str, java_options: List[str]) -> 'RunnerInterface':
         interfaces = RunnerInterface._get_interfaces()
-        matching_interfaces = [i for i in interfaces if i.version() == requested_version]
-        if matching_interfaces:
-            return matching_interfaces[0](jar_path, java_options)
-        else:
-            raise NoCompatibleRunnerInterface(requested_version)
```
salsolatragus: This error case doesn't completely go away with the new implementation, does it? Someone could specify a version that is smaller than any supported interface. Is there another case where wouldn't have a match? If this is the only case, we could maybe issue a warning and try the most recent interface?

PR #375, opened 2018-05-15
Resolve #355: Provide a ./mubench debug command

### LINKED ISSUE #355: Provide a ./mubench debug command
The command should prepare project data, such that it is available from the host system and output the command-line arguments required to run a detector on this data from the IDE.

M8is:
Resolve #355: Provide a ./mubench debug command

Example usage with DemoDetector:
`./mubench debug /mubench/mubench.cli/target/DemoDetector.jar ex2 --only aclang`

Also, the script uses the `debug` tag, which is now handled by the detector implementation. This makes sure we don't "override" any "real" detectors.

We implicitly add the `--tag debug --force-detect` and `run` to the command for a bit more convenience.

### DIFF
--- a/data/testng/versions/677302c/version.yml
+++ b/data/testng/versions/677302c/version.yml
@@ -9,5 +9,4 @@ misuses:
 - 'dmmc-4'
 - 'grouminer-4'
 - 'grouminer-17'
-- 'mudetect-15'
 revision: 677302cb8b5a2507df97c5822eef3a03ebc4e23a^1
--- a/mubench.bin/debug
+++ b/mubench.bin/debug
@@ -1,0 +1,7 @@
+SRC=$(dirname $1)
+FILENAME=$(basename -- $1)
+DNAME=${FILENAME%.*}
+TARGET="/mubench/detectors/${DNAME}/debug"
+shift
+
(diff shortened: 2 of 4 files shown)


mubench.pipeline/tests/data/test_detector.py
```diff
@@ -96,9 +96,21 @@ def test_uses_first_release_by_default(self):
         assert_equals(expected_url, detector.jar_url)
         assert_equals("-md5_requested-", detector.md5)
 
+    def test_uses_first_cli_version_for_debug_release(self):
+        self.setup_releases([
+            {"md5": "-md5_0-", "tag": "-release_0-", "cli_version": RunnerInterfaceTestImpl.TEST_VERSION},
+            {"md5": "-md5_1-", "tag": "-release_1-", "cli_version": "0.-1"}])
```
salsolatragus: I was a little surprised by this. But I guess the easiest way to tell the pipeline which CLI version to use is indeed to have a `release.yml` in the directory with the detector Jar that I want to debug... which brings me to the crucial point: we need to update the documentation ;)

salsolatragus: Let's try to generate and mount a `release.yml` with a CLI version specified on the command line, to avoid any special handling of this case.

PR #386, opened 2018-05-18
Fix conditions in mubench script

M8is:
I noticed that with the previous version of the script, we would constantly run into help when one argument was passed (e.g. `./mubench browse`), because the `&&` conditions didn't work correctly.

I'm not sure if this is the correct way to do this, but it works for me.

### DIFF
--- a/mubench
+++ b/mubench
@@ -11,7 +11,7 @@ if [[ "$*" =~ "-h" ]]; then
   HELP=t
 fi
 
-if [ $HELP ] && [ $# = 1 ]; then
+if (( $HELP )) && (( $# -eq 1 )); then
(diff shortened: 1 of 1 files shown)


salsolatragus:
Somehow this is still broken for me... on `master` when I type `./mubench browse` nothing happens and when I type `./mubench browse -h` I get the help. On this PR, when I type either command, I get an interactive shell...

M8is:
Weird... the problem seems to be that the `&&` if conditions aren't evaluated correctly. Thought I fixed it, but I probably just moved the problem to the inner if statement. I'll check it again.

M8is:
Got it now. The problem was that the check `if [ $HELP ] && [ $# = 1 ]` checks if `$HELP` is set. Since this was true either way (even if `$HELP = false`), we always evaluated that to true and only checked the other half of the statement. 
There are two options to fix this: use `[ $HELP = true ]`, which I find a bit clunky, or just don't set `$HELP` if no help flag is present, which we do now.

PR #388, opened 2018-05-21
Resolve #319: Update run modified date on new findings

### LINKED ISSUE #319: Update run modified date on new findings
I would like the run modification date to show the time that the latest findings for the run was inserted. Should be possible using Eloquents "touches" property: https://laravel.com/docs/5.0/eloquent#touching-parent-timestamps

joschli:


mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -488,7 +488,9 @@ private function storeFindings(Detector $detector, Experiment $experiment, $proj
             if(!$misuse){
                 $misuse = $this->createMisuse($detector, $experiment, $projectId, $versionId, $misuseId, $run->id);
             }
-            $this->storeFinding($detector, $experiment, $projectId, $versionId, $misuseId, $misuse, $finding);
+            $stored_finding = $this->storeFinding($detector, $experiment, $projectId, $versionId, $misuseId, $misuse, $finding);
+            $run->updated_at = $stored_finding->created_at;
```
salsolatragus: Update once, after the loop.

mubench.reviewsite/tests/RunsControllerTest.php
```diff
@@ -407,6 +407,48 @@ function test_get_misuse_results_no_runs()
...
+                [
+                    "misuse" => "-m-",
+                    "rank" => 0,
+                    "target_snippets" => [
+                        ["first_line_number" => 5, "code" => "-code-"]
+                    ]
+                ]
+            ]
+        ]);
+        sleep(1);
```
salsolatragus: Instead of waiting, can we manually change the date and expect this change to be overwritten?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -488,11 +488,14 @@ private function storeFindings(Detector $detector, Experiment $experiment, $proj
...
                 $misuse = $this->createMisuse($detector, $experiment, $projectId, $versionId, $misuseId, $run->id);
             }
-            $this->storeFinding($detector, $experiment, $projectId, $versionId, $misuseId, $misuse, $finding);
+            $last_finding = $this->storeFinding($detector, $experiment, $projectId, $versionId, $misuseId, $misuse, $finding);
+
             if ($experiment->id === 2) {
                 $this->storeFindingTargetSnippets($detector, $projectId, $versionId, $misuseId, $finding['file'], $finding['target_snippets']);
             }
         }
+        $run->updated_at = $last_finding->created_at;
```
salsolatragus: Doesn't this crash if there are not findings? There should be a test covering this case. Please write one, if there isn't.

joschli: This method is called only if there are findings.

PR #394, opened 2018-05-28
Resolve #349: Write Selenium tests for review site

### LINKED ISSUE #349: Write Selenium(?) tests for review site

joschli:


salsolatragus:
I'm not sure what's the state of this, should I review something?

joschli:
You can review it, the selenium tests now run in the ci container and i added tests for the detector and review page.

mubench.reviewsite/tests/DetectorTemplateTest.php
```diff
@@ -0,0 +1,59 @@
...
+
+    public function setUp()
+    {
+        parent::setUp();
+        $this->getRoute("/experiments/1/detectors/TestDetector/runs");
+    }
+
+    public function testAssertExperimentHeadline()
+    {
+        $this->getRoute("/experiments/1/detectors/TestDetector/runs");
```
salsolatragus: Redundant

mubench.reviewsite/tests/DetectorTemplateTest.php
```diff
@@ -0,0 +1,59 @@
...
+        );
+        self::assertFalse($modal->isDisplayed());
+        $commentIcon->click();
+        self::assertTrue($modal->isDisplayed());
+    }
+
+    public function testOpenFirstMisuseReview()
+    {
+        $viewLinks = $this->driver->findElements(
+            WebDriverBy::cssSelector("td > a.button")
```
salsolatragus: Query first table row -> query button with text "view" -> click

shippable.yml
```diff
@@ -24,6 +24,11 @@ script:
   # mubench.reviewsite
   - cd mubench.reviewsite
   - ./configure.sh
+  - java -jar /usr/local/bin/selenium-server-standalone-3.12.0.jar &
+  - cp settings.default.php settings.php
+  - touch test.sqlite
+  - php -S localhost:8080 -t ./ &
+  - sleep 10
   - phpunit --bootstrap vendor/autoload.php tests/
```
salsolatragus: Separate selenium tests from unit tests, group them in two [test suites](https://stackoverflow.com/questions/38588029/excluding-certain-tests-from-loading-in-phpunit), and execute both/the full suite on shippable.

shippable.yml
```diff
@@ -24,6 +24,11 @@ script:
   # mubench.reviewsite
   - cd mubench.reviewsite
   - ./configure.sh
+  - java -jar /usr/local/bin/selenium-server-standalone-3.12.0.jar &
+  - cp settings.default.php settings.php
+  - touch test.sqlite
```
salsolatragus: Do cp and touch before starting the servers.

docker/Dockerfile_ci
```diff
@@ -1,84 +1,113 @@
...
+            chromium-browser \
+            chromium-chromedriver \
       #Install selenium
-      && wget -q https://selenium-release.storage.googleapis.com/3.12/selenium-server-standalone-3.12.0.jar \
-      && mv selenium-server-standalone-3.12.0.jar /usr/local/bin/selenium-server-standalone-3.12.0.jar
+            && wget -q https://selenium-release.storage.googleapis.com/3.12/selenium-server-standalone-3.12.0.jar \
+          && mv selenium-server-standalone-3.12.0.jar /usr/local/bin/selenium-server-standalone-3.12.0.jar \
+          && ln /usr/lib/chromium-browser/chromedriver /usr/local/bin/chromedriver
+
+# Cleanup
```
salsolatragus: This does not make sense. It does not make the image smaller, but larger! Every Docker command creates a new layer with the diff to the underlying layers. If you want to remove things, execute a cleanup at the end of a command that installs stuff. Then the cleanup happens before the diff is generated and the diff might actually get smaller.

docker/Dockerfile_ci
```diff
@@ -1,84 +1,113 @@
...
+# Install java8
+RUN add-apt-repository ppa:webupd8team/java \
+      && apt-get update \
+      && echo oracle-java8-installer shared/accepted-oracle-license-v1-1 select true | debconf-set-selections \
+      && apt-get install -y oracle-java8-installer
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+ENV PATH $PATH:$JAVA_HOME/bin
+
+RUN update-alternatives --set java "$JAVA_HOME/jre/bin/java" \
```
salsolatragus: You should be able to set the ENV before installing Java, which allows you to merge the two RUNs into one and make the image smaller.

mubench.reviewsite/phpunit.xml
```diff
@@ -0,0 +1,13 @@
+<phpunit bootstrap="vendor/autoload.php"
+         defaultTestSuite="unit_tests">
+    <testsuites>
+        <testsuite name="integration_tests">
+            <directory>tests/selenium</directory>
```
salsolatragus: If you place the Selenium tests below the unit tests, doesn't IntelliJ still try to execute them when running "all" tests?

mubench.reviewsite/run.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
```
salsolatragus: This script should be called `run_all_tests` or something. `run` does not really reflect its behaviour.

mubench.reviewsite/run.sh
```diff
@@ -0,0 +1,5 @@
...
+
+java -jar /usr/local/bin/selenium-server-standalone-3.12.0.jar &> /dev/null &
+php -S localhost:8080 -t /mubench/mubench.reviewsite &> /dev/null &
+phpunit --bootstrap vendor/autoload.php tests/
```
salsolatragus: Is there no need to wait for server startup here like in the CI environment?

mubench.reviewsite/tests/selenium/IndexTemplateTest.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+use Facebook\WebDriver\WebDriverBy;
+
+require_once 'WebDriverTestCase.php';
+
+class IndexTemplateTest extends WebDriverTestCase
```
salsolatragus: I don't like that these are called `XTemplateTest`. They're not specifically testing the templates, but the entire review site through the UI, which happens to be generated from a template. I think we should try to make these real scenario-based tests, like `ReviewFindingsTest`, `ManageRunsTest`, `ViewDetectorRunsTest`, ... probably most of the tests then happen to involve mainly a single template, but that doesn't define them.

mubench.reviewsite/tests/selenium/IndexTemplateTest.php
```diff
@@ -0,0 +1,46 @@
+<?php
+
+use Facebook\WebDriver\WebDriverBy;
+
+require_once 'WebDriverTestCase.php';
```
salsolatragus: Why isn't this in a namespace from which it can be imported through `use`?

docker/Dockerfile_ci
```diff
@@ -7,7 +7,10 @@ ENV DEBIAN_FRONTEND noninteractive
 ENV LANG=C.UTF-8 \
     GRADLE_VERSION=4.0.2 \
     GRADLE_HOME=/usr/local/gradle
+ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+ENV PATH $PATH:$JAVA_HOME/bin
```
salsolatragus: Does `$GRADLE_HOME` need to be on the `$PATH`?

mubench.reviewsite/tests/selenium/ViewDetectorRunsTestTest.php
```diff
@@ -1,10 +1,10 @@
 <?php
 
-use Facebook\WebDriver\WebDriverBy;
+namespace MuBench\ReviewSite\Tests\Selenium;
 
-require_once 'WebDriverTestCase.php';
+use Facebook\WebDriver\WebDriverBy;
 
-class DetectorTemplateTest extends WebDriverTestCase
+class ViewDetectorRunsTestTest extends WebDriverTestCase
```
salsolatragus: TestTest :)

PR #396, opened 2018-05-28
Resolve #370: Detector-specific snippets are not deleted with detector runs

### LINKED ISSUE #370: Detector-specific snippets are not deleted with detector runs
When I publish a run with findings and snippets, delete this run, and publish the same run with different snippets (e.g., because I fixed a bug in the snippets extraction), the review site shows both the old and the new snippet, because snippets are not deleted with the run. We should delete detector-specific snippets with the respective findings.

joschli:


mubench.reviewsite/tests/RunsControllerTest.php
```diff
@@ -469,4 +480,13 @@ private function someValidRunRequestBody()
             ]
         ];
     }
+
+    private function addSimpleMetadata($misuseId = '-m-')
```
salsolatragus: Make project and version parameters here.

mubench.reviewsite/src/Models/Misuse.php
```diff
@@ -45,6 +45,9 @@ public function findings()
     public function snippets()
     {
         $snippets = Snippet::of($this->getProject(), $this->getVersion(), $this->misuse_muid, $this->getFile())->get();
+        if($this->run->experiment_id == 2){
+            $snippets = $snippets->where('detector_muid', $this->detector->muid);
```
salsolatragus: Instead of checking for Ex2, can we query by `(detector_id = $this->detector->muid OR detector_id IS NULL)`?

mubench.reviewsite/src/Controllers/RunsController.php
```diff
@@ -214,6 +215,11 @@ function deleteRunAndRelated(Run $run, $detectorId)
             $misuse->reviews()->delete();
             $misuse->metadata()->dissociate();
             $misuse->misuse_tags()->detach();
+            if($run->experiment_id == 2){
```
salsolatragus: Instead of checking for Ex2, can we delete conditionally by `detector_id IS NOT NULL`?

PR #402, opened 2018-06-25
Fix #393: NPE prevents reviewing further findings

### LINKED ISSUE #393: NPE prevents reviewing further findings
When I upload 22 findings to a review site (ex2) that has `default_ex2_review_size = 20`, I cannot review finding 21, because there's a NPE

    [2018-05-24 14:04:03] mubench.CRITICAL: Error: Call to a member function getReview() on null in .../src/Controllers/ReviewsController.php:51

The NPE occurs [here](https://github.com/stg-tud/MUBench/blob/master/mubench.reviewsite/src/Controllers/ReviewsController.php#L51). It's [the logic that determines the next/current/previous misuse](https://github.com/stg-tud/MUBench/blob/master/mubench.reviewsite/src/Controllers/ReviewsController.php#L163) that returns `null` for the current misuse. Interestingly, when I set `default_ex2_review_size = 21`, I can successfully review finding 21, but I (still) cannot review finding 22, so somehow the default impacts the querying here.

joschli:


mubench.reviewsite/src/Controllers/ReviewsController.php
```diff
@@ -41,10 +41,24 @@ public function getReview(Request $request, Response $response, array $args)
...
+            $all_misuses = $this->collectAllMisuses($runs);
+
+            list($previous_misuse, $next_misuse, $next_reviewable_misuse, $misuse) =
+                $this->determineNavigationTargets($all_misuses, $project_muid, $version_muid, $misuse_muid, $reviewer);
+
+            $next_misuse = $all_misuses->first();
+            if($limited_all_misuses->where('id', $next_reviewable_misuse->id)->isEmpty()){
+                $next_reviewable_misuse = null;
+            }
+        }
```
salsolatragus: Try to reimplement `determineNavigationTargets` by starting from the current misuse and walking forward/backward to determine the targets.

PR #444, opened 2020-05-09
Update ArgoUML url

oneturkmen:
ArgoUML's old web site seems to be down:

image

However, they seemed to have migrated to [GitHub](https://github.com/argouml-tigris-org/argouml). 

This PR updates the link from the old svn repo to the new git repo of ArgoUML.

### DIFF
--- a/data/argouml/project.yml
+++ b/data/argouml/project.yml
@@ -1,5 +1,5 @@
 name: Argouml
 repository:
-  type: svn
-  url: http://argouml.tigris.org/svn/argouml/trunk
-url: http://argouml.tigris.org/
+  type: git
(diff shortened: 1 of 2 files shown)


salsolatragus:
@oneturkmen Thanks for submitting this pull request! Since this migration include a move from SVN to Git, we also need to adapt the `revision` in `data/argouml/versions/026/version.yml`. Therefore, we need to identify the git commit that corresponds to the former SVN revision Id 15810. Could you take care of that?

oneturkmen:
Found it and changed the corresponding file.

image

salsolatragus:
Thanks a lot. I'll try to test the change over the weekend.

salsolatragus:
Works fine. Thanks again!

PR #452, opened 2021-07-02
Minor bug fix

DehengYang:
Ensure "APT::Get::Assume-Yes \"true\";\nAPT::Get::force-yes \"true\";" is correctly written to /etc/apt/apt.conf.d/90forceyes


salsolatragus:
Thank you very much, @DehengYang, for figuring this out!

